PR #1078, opened 2021-04-15
chore(): fix the case of empty providers

SeverS:
## Checklist

### DIFF
--- a/ui/hooks/src/use-profile.ts
+++ b/ui/hooks/src/use-profile.ts
@@ -409,8 +409,6 @@ export const useProfile = (
       actions.resetUpdateStatus();
     },
     getUsernameTypes() {
-      // default username is the one set in
-      // the onboarding flow (UsernameTypes.TEXTILE)
       const type: UsernameTypes[] = [];
 
       const defaultProvider = profile.default?.find(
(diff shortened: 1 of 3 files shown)


ui/hooks/src/use-profile.ts
```diff
@@ -440,6 +438,23 @@ export const useProfile = (
...
+       * in the onboarding phase
+       */
+      if (profile.providers?.length === 0 && profile.userName) {
+        const call = props.profileService.addProfileProvider({
+          property: ProfileProviderProperties.USERNAME,
+          value: profile.userName,
+          provider: ProfileProviders.EWA_BASIC,
+        });
+        call.subscribe(() => {
+          // nothing to do here as we don't really
```
kenshyx: there should be an error handler here, also log the error message just in case

ui/hooks/src/use-profile.ts
```diff
@@ -448,10 +449,17 @@ export const useProfile = (
...
-        call.subscribe(() => {
-          // nothing to do here as we don't really
-          // need to wait for the response
-        });
+        call.subscribe(
+          () => {
+            // nothing to do here as we don't really
+            // need to wait for the response
+          },
+          (err: Error) => {
```
kenshyx: let's not rely on order, it is better to be specific:
```
.subscribe({
  next(x) { console.log('got value ' + x); },
  error(err) { logger.error(`Error in getUsernameTypes: ${err}`); },
})
```

kenshyx (approved):

PR #1080, opened 2021-04-15
Feat: implement anchors

josenriagu:
## Description

- use anchors to enable opening links in new tabs

## Checklist


apps/akasha/src/components/feed-page/entry-card-renderer.tsx
```diff
@@ -151,6 +151,8 @@ const EntryCardRenderer = (props: IEntryCardRendererProps) => {
                   style={{ height: 'auto', ...style }}
                   bookmarkLabel={t('Save')}
                   bookmarkedLabel={t('Saved')}
+                  profileAnchorLink={'/profile'}
+                  repliesAnchorLink={'/social-app/post'}
```
kenshyx: @quininez would it make sense to use the post route from the akasha-app, or we leave it like this?

kenshyx: @josenriagu this link value is available here https://github.com/AKASHAorg/akasha-world-framework/blob/next/apps/akasha/src/routes.tsx#L14

quininez: Right now we can't get the /profile link from a constant so this will need changing in the future anyway, best to also leave a todo comment there, but I would import the link for /post from the routes file now

josenriagu: the routes can only be accessed from apps, so I will change those being passed from apps

kenshyx (approved):

PR #1102, opened 2021-04-20
Moderation API migration (ready for review)

deiu:
This PR adds the moderation API that was previously written in Go.

## Issues that will be closed

TBD

## Testing

TBD

## Checklist


api/hub/.env.example
```diff
@@ -23,4 +25,6 @@ AWF_FAUCET_NETWORK=
 
 AWF_MAILSENDER_KEY=
 
+AWF_MODERATION_ADMIN_SECRET=
```
kenshyx: this one looks like it's not used

api/hub/src/api.ts
```diff
@@ -27,26 +64,295 @@ api.post('/validate-token/:token', async (ctx: koa.Context, next: () => Promise<
...
+
+api.post('/moderation/decisions/pending', async (ctx: koa.Context, next: () => Promise<any>) => {
+  const req = ctx?.request.body;
+  const decisions = await dataSources.decisionsAPI.listDecisions(
+    false,
+    false,
+    req.offset,
+    req.limit,
+  );
+  const list = await Promise.all(
```
kenshyx: replace this with `for of`

api/hub/.env.example
deiu: You're right, I've removed it now.

api/hub/src/api.ts
kenshyx: something like
```
const list = [];
for(const decision of decisions){
  const result = await getFinalDecision(decision.contentID);
  list.push(result);
}
```

kenshyx: I see this block reused, it should be updated everywhere

api/hub/src/api.ts
```diff
@@ -27,26 +64,295 @@ api.post('/validate-token/:token', async (ctx: koa.Context, next: () => Promise<
...
+      if (!profile || !profile.pubKey || !request.data.signature) {
+        ctx.status = 401;
+      } else {
+        try {
+          const verified = await verifyEd25519Sig({
+            pubKey: profile.pubKey,
+            data: request.data,
+            signature: request.signature,
+          });
+          const isAdmin = dataSources.moderatorsAPI.isAdmin(request.data.adminEthAddress);
```
kenshyx: I think here you should have `await` or it will be always `true` because the result is a Promise

deiu: Good catch!

api/hub/src/datasources/moderation-decision.ts
```diff
@@ -0,0 +1,88 @@
...
+      .and('moderated')
+      .eq(moderated)
+      .skipNum(offset || 0)
+      .limitTo(limit || 10);
+    return db.find<ModerationDecision>(this.dbID, this.collection, query);
+  }
+
+  async getDecision(contentId: string) {
+    const db: Client = await getAppDB();
+    const query = new Where('contentID').eq(contentId);
```
kenshyx: this is very slow, you can add a caching strategy to resolve faster subsequently calls

deiu: I agree.

kenshyx:
Also the base for this PR should be `refactor-v0.2`

api/hub/src/api.ts
```diff
@@ -236,11 +220,11 @@ api.post('/moderation/decisions/pending', async (ctx: koa.Context, next: () => P
...
     req.limit,
   );
-  const list = await Promise.all(
-    decisions.map(decision => {
-      return getFinalDecision(decision.contentID);
-    }),
-  );
+  const list = [];
+  for (const decision of decisions) {
+    console.log('\n\nPENDING:\n\n', decision);
```
kenshyx: replace console logs with logger from helpers

deiu: I meant to remove it but forgot. It's gone now.

api/hub/src/datasources/moderation-report.ts
```diff
@@ -0,0 +1,162 @@
...
+    if (exists) {
+      return Promise.reject(`cannot report content ${contentID} twice as user ${author}`);
+    }
+
+    // Do we need to check if a pending decision was created for this post?
+    const query = new Where('contentID').eq(contentID);
+    const results = await db.find<ModerationReport>(this.dbID, decisionsCollection, query);
+    if (results.length < 1) {
+      // need to create a pending decision to track moderation status
+      // how to avoid race condition here?
```
kenshyx, line 170: you can have an `unique` index on `contentID` so there is just one decision record :)

kenshyx: but I saw that this is already done so what's the remaining issue here?

api/hub/src/datasources/moderation-report.ts
```diff
@@ -0,0 +1,162 @@
...
+    const reports = await db.find<ModerationReport>(this.dbID, this.collection, query);
+    const list = [];
+    for (const report of reports) {
+      report.explanation = decodeString(report.explanation);
+      list.push(report);
+    }
+    return list;
+  }
+
+  async getReport(contentID: string, author: string) {
```
kenshyx, line 83: some comments on each method in the ts-doc format would be nice :D

deiu: You are right. I'll add these before merging.

api/hub/src/datasources/moderation-report.ts
deiu: That is actually the case. I'm think there might be a possible race condition when two users report the same content and a pending decision is created. Unless `db.create` fails graciously in case a record exists. I can also not throw in case of an error there.

kenshyx: it will throw automatically on the 2nd insert because of the unique constraint

deiu: I know, but it should be ignored since we don't care about it on the 2nd insert. Would removing the throw (a few lines below) just skip over this part of the code?

kenshyx: when the error happens it will automatically also return a response

deiu:
I've added comments. Ready for final review.

api/hub/src/datasources/moderation-decision.ts
```diff
@@ -0,0 +1,185 @@
...
+   * @returns string
+  */
+  getCountersCacheKey() {
+    return `${this.collection}:counters`;
+  }
+
+  /**
+   * Count all decisions (pending, moderated and delisted, moderated and kept).
+   * @returns Object
+   * ```typescript
```
kenshyx, line 48: you can also add `@example` tag here, this is compliant with tsdoc

api/hub/src/datasources/moderation-admin.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (results.length) {
+      return results[0];
+    }
+    logger.warn(`could not find moderator with address ${ethAddress}`);
+    return undefined;
+  }
+
+  /**
+   * Check if the given user is a moderator.
+   * @param ethAddress the ETH address of the moderator
```
kenshyx: the `@param comment` format for this should be `@param ethAddress - The ETH address of the moderator`

kenshyx: you can test out [here](https://tsdoc.org/play#) to see how it looks like

api/hub/src/datasources/moderation-admin.ts
```diff
@@ -23,7 +23,7 @@ class ModerationAdminAPI extends DataSource {
 
   /**
    * Get data for a moderator.
-   * @param address the ETH address of the moderator
+   * @param address The ETH address of the moderator
```
kenshyx: it should also contain the `-` before the description like `@param address - The ETH address of the moderator`

deiu: Done, for some reason I was following the TypeDoc standard instead of tsdoc.

kenshyx (approved):

PR #1157, opened 2021-05-27
Email notifications for moderation

deiu:
Added email notifications for moderation.

## Description

This PR implements a helper function that uses SendGrid to send of emails. In this instance we are using it to send email notifications every time a new moderation request is made (i.e. a user has reported content). Emails are sent **only once** per content identifier that was reported -- i.e. a second user reporting the same content will **NOT** trigger a new email to be sent.

## Checklist


deiu:
Ready for review.

api/hub/src/datasources/moderation-report.ts
```diff
@@ -195,6 +195,26 @@ class ModerationReportAPI extends DataSource {
...
+      const text = `There is a new pending request for moderation.
+      Please visit the moderation app (https://akasha.ethereum.world/moderation-app/home)
+      to moderate this content.
+      \nThank you!`;
+      const html = `<p>There is a new pending request for moderation.
+      Please visit the <a href="https://akasha.ethereum.world/moderation-app/home">moderation app</a>
+      to moderate this content.</p>
+      <br>
+      <p>Thank you!</p>`;
+      try {
```
kenshyx: this should have a debounce period like 1h, or we end up with sending emails for every report made

api/hub/src/datasources/moderation-report.ts
```diff
@@ -195,6 +195,26 @@ class ModerationReportAPI extends DataSource {
...
+        await sendEmailNotification(process.env.MODERATION_EMAIL,
```
kenshyx: if you need to try catch then you can use directly the promise api `sendEmailNotification(...).then(_=>whatever).catch(err => do err stuff)`

api/hub/src/helpers.ts
```diff
@@ -209,6 +210,32 @@ export const sendAuthorNotification = async (
...
+ * @returns A promise that resolves upon sending the email
+ */
+export const sendEmailNotification = async (
+to: string,
+from: string,
+subject: string,
+text: string,
+html: string
+) => {
+  sendgrid.setApiKey(process.env.SENDGRID_API_KEY)
```
kenshyx: you dont need to set this on every send call, set it up outside

api/hub/src/helpers.ts
```diff
@@ -209,6 +210,32 @@ export const sendAuthorNotification = async (
...
+ * Send an email notification using SendGrid.
+ * @param to - Destination email address
+ * @param from - Source email address
+ * @param subject - Email subject
+ * @param text - Email body (text version)
+ * @param html - Email body (HTML version)
+ * @returns A promise that resolves upon sending the email
+ */
+export const sendEmailNotification = async (
+to: string,
```
kenshyx: consider using an object param with these attributes, makes the function signature easier to read when it is called

api/hub/src/datasources/moderation-report.ts
deiu: Actually, this email is only sent on the first report for each content ID (when the pending moderation item is created). All subsequent reports will not trigger it.

api/hub/src/datasources/moderation-report.ts
deiu: Done.

kenshyx (approved):

PR #1190, opened 2021-06-11
Avoid issues if moderator is undefined

deiu:
Fixes undefined result when checking if user is moderator.

## Description

Added a check to avoid undefined moderator object when checking for moderator status.

## Checklist

- [] I have added tests to cover my changes

### DIFF
--- a/api/hub/src/datasources/moderation-admin.ts
+++ b/api/hub/src/datasources/moderation-admin.ts
@@ -59,7 +59,7 @@ class ModerationAdminAPI extends DataSource {
    */
   async isModerator(ethAddress: string) {
     const moderator = await this.getModerator(ethAddress);
-    return moderator.active;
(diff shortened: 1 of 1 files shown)


api/hub/src/datasources/moderation-admin.ts
```diff
@@ -59,7 +59,7 @@ class ModerationAdminAPI extends DataSource {
    */
   async isModerator(ethAddress: string) {
     const moderator = await this.getModerator(ethAddress);
-    return moderator.active;
+    return moderator && moderator.active;
```
kenshyx, line 62: if active is optional then this can be 

```
return !!moderator?.active
```

kenshyx: this works also

kenshyx (approved):

PR #1212, opened 2021-07-01
Feat: transparency log

josenriagu:
## Description

-  composes transparency log page in moderation app

## Checklist


apps/moderation/src/components/transparency-log.tsx
```diff
@@ -0,0 +1,293 @@
...
+  };
+
+  const buttonValues = !isMobile
+    ? ['All', 'Kept', 'Delisted']
+    : ['All', 'Kept', 'Delisted', 'Stats'];
+  const buttonLabels = !isMobile
+    ? [t('All'), t('Kept'), t('Delisted')]
+    : [t('All'), t('Kept'), t('Delisted'), t('Stats')];
+
+  const baseAvatarUrl = 'https://hub.textile.io/ipfs/';
```
quininez: Why is this defined as a constant here? this url should come from the sdk, like: sdk.services.common.ipfs.getSettings().gateway

apps/moderation/src/components/transparency-log.tsx
```diff
@@ -0,0 +1,293 @@
...
+          `${selected.contentType.charAt(0).toUpperCase()}${selected.contentType.substring(1)} ${
+            selected.delisted ? 'delisted by' : 'kept by'
+          }`,
+        )}
+        content={t(`${selected.explanation}`)}
+        isDelisted={selected.delisted}
+        moderator={selected.moderator.name}
+        moderatedTimestamp={selected.moderatedDate.toString()}
+        moderatorAvatarUrl={
+          baseAvatarUrl + selected.moderator.avatar || 'https://placebeard.it/360x360'
```
quininez: Isn't the moderator just another user profile? If you are  using the useProfile hook to get the profile data for the moderator then the avatar will already contain the full image path.

quininez: You shouldn't use https://placebeard.it/360x360 as a placeholder inside the app, this is for internal use in the storybook only, just use Avatar design system component which loads our platform specific placeholders

quininez (changes requested):

apps/moderation/src/components/transparency-log.tsx
josenriagu: gotcha 👍

apps/moderation/src/components/transparency-log.tsx
josenriagu: mapping through the list, since we have the avatar coming from the api response, the ipfsgateway link will work just fine

quininez (approved):

PR #1214, opened 2021-07-02
feat: added moderation action log

deiu:
The PR implements the actions log for moderation decisions.

## Description

The actions log is necessary to track decisions changes for any content. For example, if a decision was reverted or illegally changed.

## Checklist


kenshyx:
This collection `ModerationAction` looks like a duplicate of `ModerationDecision`. I would suggest to update the schema of `ModerationDecision` instead of doing 2x writes of the same data.

deiu:
It's actually not that simple. The issue that prevented me from doing exactly this, is that Textile doesn't have a `unique` constraint when querying the database. For example, when building the list of moderation decisions (assuming each decision change is stored in the `ModerationDecision` collection), then I want to return unique results based on the contentIDs.

The current approach does perform 2x write operations, but they are actually an update (of the final decision) and a write (in the action log). Basically if a decision changes n times, then it will be stored n+1 times in the database (n changes + 1 final decision).

kenshyx (approved):

PR #1228, opened 2021-07-16
Feat/remove entries

SeverS:
## Description

Remove posts and replies

## Checklist


quininez (approved):

apps/akasha/src/extensions/entry-remove-modal.tsx
```diff
@@ -0,0 +1,191 @@
...
+import Fetch from 'i18next-fetch-backend';
+import LocalStorageBackend from 'i18next-localstorage-backend';
+import { QueryClientProvider, QueryClient } from 'react-query';
+import getSDK from '@akashaproject/awf-sdk';
+import { events } from '@akashaproject/awf-sdk/typings';
+import { filter } from 'rxjs/operators';
+
+const { ThemeSelector, lightTheme, darkTheme, ConfirmationModal } = DS;
+
+const queryClient = new QueryClient();
```
kenshyx, line 20: you don't need this anymore if you're not using hooks from `use-posts.new`

kenshyx: you can safely remove it from package.json and the QueryClient provider

quininez: in the future it will be migrated to react query, so it still needs it, but you just need to wrap the ModalWrapper with the provider-hoc.tsx withProviders from hooks, and that gives QueryClientProvider and ThemeSelector

PR #1231, opened 2021-07-19
fix: action log backwards compatibility

deiu:
This fix makes sure there are no errors when trying to get the action for (old) moderated content that doesn't have the `actions` property.

## Checklist

### DIFF
--- a/api/hub/src/datasources/moderation-decision.ts
+++ b/api/hub/src/datasources/moderation-decision.ts
@@ -132,7 +132,7 @@ class ModerationDecisionAPI extends DataSource {
    */
   async listActions(contentID: string) {
     let decision = parse(stringify(await this.getDecision(contentID)));
-    if (decision.actions && decision.actions.length > 0) {
+    if (decision.actions && decision.actions.length) {
       decision.actions.forEach((element, index) => {
         element.explanation = decodeString(element.explanation);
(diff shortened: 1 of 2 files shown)


api/hub/src/datasources/moderation-decision.ts
```diff
@@ -132,10 +132,12 @@ class ModerationDecisionAPI extends DataSource {
    */
   async listActions(contentID: string) {
     let decision = parse(stringify(await this.getDecision(contentID)));
-    decision.actions.forEach((element, index) => {
-      element.explanation = decodeString(element.explanation);
-      decision.actions[index] = element;
-    });
+    if (decision.actions) {
```
kenshyx: ensure that it is also a list

api/hub/src/datasources/moderation-decision.ts
```diff
@@ -132,7 +132,7 @@ class ModerationDecisionAPI extends DataSource {
    */
   async listActions(contentID: string) {
     let decision = parse(stringify(await this.getDecision(contentID)));
-    if (decision.actions) {
+    if (decision.actions && decision.actions.length > 0) {
```
kenshyx: you can check only for the length prop without comparison

kenshyx (approved):

PR #1238, opened 2021-07-28
feat: also return reported reason when checking moderation status

deiu:
This PR adds an extra attribute to the data returned by the moderation API. It is required by the new design screens when checking if content was reported by the same user.

## Checklist


josenriagu (approved):
Looks good

PR #1250, opened 2021-08-17
Feat/scroll position

SeverS:
## Checklist


ui/hooks/src/use-posts.new.ts
```diff
@@ -143,16 +166,14 @@ const getPost = async postID => {
 
 // hook for fetching data for a specific postID/entryID
 export function usePost(postID: string, enabler: boolean) {
+  const queryClient = useQueryClient();
   return useQuery([ENTRY_KEY, postID], () => getPost(postID), {
     enabled: !!postID && enabler,
+    initialData: () => queryClient.getQueryData([ENTRY_KEY, postID]),
```
quininez: How does this work? Is it to load data from the cache initially if that post was previously loaded?

SeverS: yes, but this was a leftover from the initial iteration where we were storing everything inside infiniteQuery hook, i will remove it

quininez (approved):

PR #1259, opened 2021-08-24
feat: worker threads for notifications

deiu:
This PR adds support for worker threads when sending (Textile) notifications.

## Checklist


api/hub/src/worker.js
```diff
@@ -0,0 +1,5 @@
+const path = require('path');
```
kenshyx: you don't need to create this file as .js, just make it .ts and use the lib path for imports

api/hub/src/notifications.ts
```diff
@@ -0,0 +1,52 @@
...
+  const textEncoder = new TextEncoder();
+  const encodedNotification = textEncoder.encode(JSON.stringify(notificationObj));
+  return ms.sendMessage(recipient, encodedNotification);
+};
+
+sendMessage(workerData.recipient, workerData.notificationObj).then(() => {
+  parentPort.postMessage(undefined);
+}).catch( error => {
+  parentPort.postMessage(error);
+})
```
kenshyx: fix your editor settings, always add at the end of the file a newline

api/hub/package.json
```diff
@@ -64,6 +64,7 @@
     "node-abort-controller": "3.0.0",
     "object-hash": "2.2.0",
     "prom-client": "13.2.0",
+    "ts-node": "^10.2.1",
```
kenshyx: no need for this dependency

api/hub/src/notifications.ts
```diff
@@ -0,0 +1,52 @@
...
+  PrivateKey,
+  PublicKey,
+  createUserAuth,
+  Users,
+} from '@textile/hub';
+import { logger } from './helpers';
+import { parentPort, workerData } from 'worker_threads';
+
+let mailSender;
+const mailSenderIdentity = () => PrivateKey.fromString(process.env.AWF_MAILSENDER_KEY);
```
kenshyx, line 11: did you test the env values being shared here?

api/hub/src/worker.js
deiu: I've tried to do that, but passing a  .ts file to a worker thread results in error:
```
The worker script extension must be “.js” or “.mjs”. Received “.ts”
```

This workaround allows us to use .ts files for the actual worker code, which then gets loaded by the .js file.

api/hub/src/notifications.ts
deiu: Yes, they work.

api/hub/package.json
deiu: This is needed by the worker loader file (the js one).

api/hub/src/worker.js
kenshyx: the tsc compilation step would create the .js file for you

kenshyx: just make this file .ts, remove ts-node and keep everything else

kenshyx: because after the compilation step everything will happen under the lib folder where everything is already .js

api/hub/src/notifications.ts
kenshyx: this still doesn't end with a newline

kenshyx (approved):

PR #1261, opened 2021-08-25
reposting fix WIP

SeverS:
## Description

- fix reposting
- typings improvements
- general error handling improvements
- use new mention and tag search hooks in editors
- comment remove fix
- cleanups (mostly on removing the older error handlers like ErrorInfoCards) and other minor fixes

## Checklist


ui/hooks/src/use-comments.new.ts
```diff
@@ -98,14 +84,25 @@ export interface Publish_Options {
   data: DataProviderInput[];
   comment: { title?: string; tags?: string[]; postID: string };
 }
-export function useDeleteComment(commentID: string) {
+
+const deleteComment = async (commentId: string) => {
   const sdk = getSDK();
+  const resp = await lastValueFrom(sdk.api.comments.removeComment(commentId));
+  if (resp.hasOwnProperty('error')) {
+    throw new Error(resp['error']);
```
quininez, line 92: Where are these errors caught?

ui/hooks/src/utils/entry-utils.ts
```diff
@@ -27,15 +34,15 @@ function fromBinary(binary: any) {
   return result;
 }
 
-export const serializeBase64ToSlate = (base64Content: string, logger?: any) => {
-  const stringContent = atob(base64Content);
+export const decodeb64SlateContent = (base64Content: string, logger?: ILogger) => {
+  const stringContent = window.atob(base64Content);
```
quininez, line 38: Why did you use it as a method (window.atob) instead of directly?

ui/hooks/src/utils/entry-utils.ts
```diff
@@ -217,11 +169,33 @@ export const createPendingEntry = (
...
+  comment: {
+    tags: IPublishData['metadata']['tags'];
+    mentions: IPublishData['metadata']['mentions'];
+    postID: string;
+  };
+}
+
+export function buildPublishObject(data: IPublishData): EntryPublishObject;
+export function buildPublishObject(data: IPublishData, parentEntryId: string): CommentPublishObject;
+export function buildPublishObject(data: IPublishData, parentEntryId?: string) {
```
quininez, line 192: Why is this defined multiple times?

ui/hooks/src/use-comments.new.ts
SeverS: https://react-query.tanstack.com/guides/query-functions#handling-and-throwing-errors

ui/hooks/src/utils/entry-utils.ts
SeverS: to avoid deprecation warning, atob is deprecated in nodejs. i can avoid that warning explicitly using window.atob

ui/hooks/src/utils/entry-utils.ts
SeverS: it's function overloading, i am returning a different type based on that `parentEntryId` parameter.
I can remove the typings in from the actual function declaration. i've left them there because it's easier to see them in the editor's autocomplete popovers

quininez: ok, good to know:)

quininez (approved):

PR #1266, opened 2021-09-06
refactor: update moderation request

josenriagu:
## Description

- sets correct reason on entry update after reporting

## Checklist

### DIFF
--- a/apps/moderation/src/extensions/report-modal.tsx
+++ b/apps/moderation/src/extensions/report-modal.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import i18next from 'i18next';
 import ReactDOM from 'react-dom';
-import { useQueryClient } from 'react-query';
 import Fetch from 'i18next-fetch-backend';
 import singleSpaReact from 'single-spa-react';
 import DS from '@akashaproject/design-system';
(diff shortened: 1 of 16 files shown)


apps/moderation/src/extensions/report-modal.tsx
```diff
@@ -57,27 +57,27 @@ const ReportModalComponent = (props: RootComponentProps) => {
     }
   }, [activeModal]);
 
-  const updateOnSuccess = (isSuccess: boolean) => {
+  const updateOnSuccess = (isSuccess: boolean, reason: string) => {
     // this method utilises react-query to update reported state depending on contentType
     if (contentType === 'post') {
       queryClient.setQueryData<unknown>([ENTRY_KEY, activeModal.entryId], prev => ({
```
SeverS: This needs a refactor. I am not ok updating the state outside of a hook. it makes it untraceable.
I would wrap the call that is made through the `moderationRequest.clickHandler` with a `useMutation` and do the mutation directly. This way you will also benefit from the status and error

ui/hooks/src/moderation-request.ts
```diff
@@ -255,7 +255,7 @@ export default {
             throw new Error('Unable to process your request right now. Please try again later');
           }
 
-          return modalName === 'report-modal' ? callback(true) : callback();
+          return modalName === 'report-modal' ? callback(true, dataToSign.reason) : callback();
```
SeverS: I would suggest to always pass those parameters to the callback and get rid of this logic here, and remove the unnecessary modalName param.

ui/hooks/src/moderation-request.ts
```diff
@@ -255,7 +255,7 @@ export default {
...
         })
         .catch(error =>
           logger.error(`[${modalName}.tsx]: fetchRequest err %j`, error.message || ''),
```
SeverS: when you log the error don't lie about the source. this is not the `${modalName}.tsx` file. It's the moderation-request.modalClickHandler. also make sure that you pass only one parameter to the `.error` and it is a `string`

SeverS (changes requested):

ui/hooks/src/moderation-request.ts
josenriagu: gotcha.

SeverS (approved):

PR #1267, opened 2021-09-07
Fix/moderation api backwards compatibility

deiu:
This PR adds a number of fixes to the moderation API to make it backwards compatible with legacy data.

## Checklist

### DIFF
--- a/api/hub/src/datasources/moderation-decision.ts
+++ b/api/hub/src/datasources/moderation-decision.ts
@@ -261,7 +261,7 @@ class ModerationDecisionAPI extends DataSource {
     if (await queryCache.has(decisionCache)) {
       return queryCache.get(decisionCache);
     }
-    let finalDecision = await this.getDecision(contentID);
+    const finalDecision = await this.getDecision(contentID);
     // remove action log as it is not needed here
     if (finalDecision.actions) {
       delete finalDecision.actions;
@@ -271,7 +271,7 @@ class ModerationDecisionAPI extends DataSource {
       const moderator = finalDecision.moderator.startsWith('0x')
         ? await profileAPI.getProfile(finalDecision.moderator)
         : await profileAPI.resolveProfile(finalDecision.moderator);
-      finalDecision = Object.assign({}, finalDecision, {
+      Object.assign(finalDecision, {
         moderator: moderator.pubKey,
         moderatorProfile: {
           ethAddress: moderator.ethAddress, // Deprecated, to be removed
@@ -300,7 +300,7 @@ class ModerationDecisionAPI extends DataSource {
     const reports = await reportingAPI.countReports(contentID);
     // add reasons
     const reasons = await reportingAPI.getReasons(contentID);
-    finalDecision = Object.assign({}, finalDecision, {
+    Object.assign(finalDecision, {
       reports,
       reportedBy,
       reportedDate,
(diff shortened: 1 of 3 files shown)


api/hub/src/datasources/moderation-decision.ts
```diff
@@ -250,14 +261,17 @@ class ModerationDecisionAPI extends DataSource {
     if (await queryCache.has(decisionCache)) {
       return queryCache.get(decisionCache);
     }
-    const decision = await this.getDecision(contentID);
+    let finalDecision = await this.getDecision(contentID);
```
kenshyx: this can be `const` if you're not reasigning `finalDecision`

api/hub/src/datasources/moderation-decision.ts
```diff
@@ -250,14 +261,17 @@ class ModerationDecisionAPI extends DataSource {
...
+      delete finalDecision.actions;
+    }
     // add moderator data
-    if (finalDecision.moderator) {
+    if (finalDecision.moderator && finalDecision.moderator.length) {
       const moderator = finalDecision.moderator.startsWith('0x')
         ? await profileAPI.getProfile(finalDecision.moderator)
         : await profileAPI.resolveProfile(finalDecision.moderator);
+      finalDecision.moderator = moderator.pubKey;
       finalDecision = Object.assign({}, finalDecision, {
```
kenshyx: this can be just a simple Object.assign with the first paramenter being `finalDecision`

api/hub/src/datasources/moderation-decision.ts
deiu: I'm calling `Object.assign()` later to add extra fields, so I can't make it a const.

api/hub/src/datasources/moderation-decision.ts
deiu: Here I'm overwriting the `moderator` attribute to make sure it will always be the pubKey. I'm not creating a new attribute.

kenshyx: so why not use a normal assignment? :D

kenshyx: btw `Object.assign(finalDecision, {....})` does exactly what you want to achieve with the reassignment

deiu: Oh, I see. OK, I've switched to a normal assignment.

api/hub/src/datasources/moderation-decision.ts
```diff
@@ -271,8 +271,8 @@ class ModerationDecisionAPI extends DataSource {
       const moderator = finalDecision.moderator.startsWith('0x')
         ? await profileAPI.getProfile(finalDecision.moderator)
         : await profileAPI.resolveProfile(finalDecision.moderator);
-      finalDecision.moderator = moderator.pubKey;
       finalDecision = Object.assign({}, finalDecision, {
+        moderator: moderator.pubKey,
```
kenshyx, line 275: I was suggesting something like 
```
Object.assign(
    finalDecision,
    { 
        moderator: moderator.pubKey,
        moderatorProfile: { ethAddress: moderator.ethAddress, pubKey: moderator.pubKey }
    }
)
```
that would replace the entire line 274

deiu: Got it. I've pushed a fix for both comments.

kenshyx (approved):

PR #1304, opened 2021-09-22
Fix/comment edit

ghost:
## Checklist

### DIFF
--- a/ui/hooks/src/use-comments.new.ts
+++ b/ui/hooks/src/use-comments.new.ts
@@ -170,13 +170,13 @@ export function useCreateComment() {
 }
 
 // hook used for editing a comment
-export function useEditComment(commentID: string, isComment: boolean) {
+export function useEditComment(commentID: string, hasCommentData: boolean) {
   const sdk = getSDK();
   const queryClient = useQueryClient();
 
   return useMutation(
     (comment: IPublishData & { postID: string }) => {
-      if (!isComment) return;
+      if (!hasCommentData) return;
       const { postID, ...commentData } = comment;
       const publishObj = buildPublishObject(commentData, postID);
       return lastValueFrom(
--- a/ui/hooks/src/use-comments.new.ts
+++ b/ui/hooks/src/use-comments.new.ts
@@ -170,13 +170,13 @@ export function useCreateComment() {
 }
 
 // hook used for editing a comment
-export function useEditComment(commentID: string, hasCommentData: boolean) {
+export async function useEditComment(commentID: string, hasCommentData: boolean) {
   const sdk = getSDK();
   const queryClient = useQueryClient();
 
   return useMutation(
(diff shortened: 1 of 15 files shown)


ui/hooks/src/use-comments.new.ts
```diff
@@ -170,12 +170,13 @@ export function useCreateComment() {
...
 
 // hook used for editing a comment
-export function useEditComment(commentID: string) {
+export function useEditComment(commentID: string, isComment: boolean) {
   const sdk = getSDK();
   const queryClient = useQueryClient();
 
   return useMutation(
     (comment: IPublishData & { postID: string }) => {
+      if (!isComment) return;
```
kenshyx: The use case for this flag is to prevent posting empty data?

ghost: Yes, throws errors in the console otherwise

kenshyx: I would suggest to rename this into something more verbose for the use case, like `hasCommentData`

ui/hooks/src/use-comments.new.ts
```diff
@@ -170,12 +170,13 @@ export function useCreateComment() {
 }
 
 // hook used for editing a comment
-export function useEditComment(commentID: string) {
+export function useEditComment(commentID: string, isComment: boolean) {
   const sdk = getSDK();
   const queryClient = useQueryClient();
 
   return useMutation(
     (comment: IPublishData & { postID: string }) => {
```
kenshyx: this function should be `async`

ui/hooks/src/use-comments.new.ts
kenshyx: the return should be consistent, `return Promise.resolve()`

ui/hooks/src/use-comments.new.ts
kenshyx: the mutation function, not the hook :D

kenshyx:
```
return useMutation(
    async (comment: IPublishData & { postID: string }) => {
 ```

File: ui/hooks/src/use-comments.new.ts [hunk 3]
```diff
@@ -170,13 +170,13 @@ export function useCreateComment() {
...
 // hook used for editing a comment
-export function useEditComment(commentID: string, hasCommentData: boolean) {
+export async function useEditComment(commentID: string, hasCommentData: boolean) {
   const sdk = getSDK();
   const queryClient = useQueryClient();
 
   return useMutation(
     (comment: IPublishData & { postID: string }) => {
-      if (!hasCommentData) return;
+      if (!hasCommentData) return Promise.resolve(null);
```
kenshyx: You can just use Promise.resolve() which will resolve to `undefined` value like `return;`

kenshyx (approved):

PR #1320, opened 2021-09-24
Refactor: moderation request

josenriagu:
## Description

- use ethAddress as fallback for new profiles
- use react query in moderation requests
- update components

## Checklist


apps/moderation/src/components/transparency-log/banner.tsx
```diff
@@ -22,10 +22,10 @@ const Banner: React.FC<IBanner> = props => {
       content={t(
         'Here you will find the moderated posts, replies, and accounts of Ethereum World. We do not reveal any personal information of the author or submitter(s) to protect their privacy.',
       )}
-      keptCount={count.kept}
+      keptCount={count?.kept}
       keptCountLabel={t('kept')}
       totalCountLabel={t('total')}
-      delistedCount={count.delisted}
+      delistedCount={count?.delisted}
```
kenshyx: I think you want here a number, not `null` value.
`{count?.delisted || 0}`

apps/moderation/src/components/transparency-log.tsx
```diff
@@ -119,7 +82,7 @@ const TransparencyLog: React.FC<ITransparencyLogProps> = props => {
     <Box>
       <SwitchCard
         count={
-          activeButton === 'All' ? count.kept + count.delisted : count[activeButton.toLowerCase()]
+          activeButton === 'All' ? count?.kept + count?.delisted : count[activeButton.toLowerCase()]
```
quininez: here it would be a better approach to create an enum with the possible values of the buttons and use that instead of text values, ex.: enum ButtonValues { ALL: 'all', KEPT: 'kept'} and use ButtonValues.ALL instead of 'All' directly

apps/moderation/src/components/transparency-log.tsx
```diff
@@ -137,56 +111,65 @@ const TransparencyLog: React.FC<ITransparencyLogProps> = props => {
...
-                    onClickAvatar={handleClickAvatar}
-                    onClickCard={handleClickCard(el)}
-                  />
-                ))
+              // map through pages, for each page, filter results according to activeButton, then map through to render the mini card
+              logItemPages.map((page, index) => (
+                <Box key={index} flex={false}>
+                  {page.results
+                    .filter((el: { delisted: boolean }) =>
+                      activeButton === 'Kept'
```
quininez: also here, replace with enum values everywhere you have text values

apps/moderation/src/components/transparency-log.tsx
```diff
@@ -137,56 +110,65 @@ const TransparencyLog: React.FC<ITransparencyLogProps> = props => {
...
+                        : el,
+                    )
+                    .map((el: ILogItem, index: number) => (
+                      <TransparencyLogMiniCard
+                        key={index}
+                        locale="en"
+                        title={t(
+                          `${el.contentType.charAt(0).toUpperCase()}${el.contentType.substring(
+                            1,
+                          )} ${el.delisted ? 'Delisted' : 'Kept'}`,
```
quininez: same here

quininez (approved):

PR #1338, opened 2021-09-28
fetch profiles from cache, fix editor typings and comment click

kenshyx:
## Checklist

### DIFF
--- a/ui/hooks/src/use-posts.new.ts
+++ b/ui/hooks/src/use-posts.new.ts
@@ -203,8 +203,8 @@ export function useDeletePost(postID: string) {
       });
       return { previousPost };
     },
-    onSuccess: () => {
-      queryClient.invalidateQueries([ENTRY_KEY, postID]);
+    onSuccess: async () => {
+      await queryClient.invalidateQueries([ENTRY_KEY, postID]);
     },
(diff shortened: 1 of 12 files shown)


ui/hooks/src/use-posts.new.ts
```diff
@@ -203,8 +203,8 @@ export function useDeletePost(postID: string) {
       });
       return { previousPost };
     },
-    onSuccess: () => {
-      queryClient.invalidateQueries([ENTRY_KEY, postID]);
+    onSuccess: async () => {
```
quininez, line 206: why make it async? I see in the docs it's used sync https://react-query.tanstack.com/guides/invalidations-from-mutations

kenshyx: the docs might be outdated, check the signature from https://github.com/tannerlinsley/react-query/blob/a53f6325e9761b1435b85cfbd337ea1a75584dde/src/core/queryClient.ts#L241

kenshyx: `queryClient.invalidateQueries('key')` returns a `Promise`

quininez (approved):

PR #1391, opened 2021-10-13
Fix: moderation app updates

josenriagu:
## Description

- removes block label from profile drop down
- updates profile card props in  moderation app
- decomposes useModeration hook

## Checklist


ui/hooks/src/use-moderation.ts
```diff
@@ -71,7 +74,63 @@ const createModerationMutation = async ({ dataToSign, contentId, contentType, ur
...
+          key.startsWith(PENDING_CACHE_KEY_PREFIX) ||
+          key.startsWith(MODERATED_CACHE_KEY_PREFIX) ||
+          key.startsWith(MODERATION_COUNT_CACHE_KEY_PREFIX)
+        ) {
+          uiCache.delete(key);
+        }
+      }
+
+      // refetch queries
+      queryClient.refetchQueries(PENDING_ITEMS_KEY);
```
kenshyx: these are async methods

ui/hooks/src/use-moderation.ts
```diff
@@ -71,7 +74,63 @@ const createModerationMutation = async ({ dataToSign, contentId, contentType, ur
...
 function useModeration() {
   const queryClient = useQueryClient();
+  const sdk = getSDK();
   return useMutation((param: UseModerationParam) => createModerationMutation(param), {
+    onSuccess: () => {
+      // remove keys from cache
+      const uiCache = sdk.services.stash.getUiStash();
+      for (const key of uiCache.keys()) {
+        if (
+          key.startsWith(PENDING_CACHE_KEY_PREFIX) ||
```
kenshyx: you should guard for undefined or null keys

ui/hooks/src/use-moderation.ts
```diff
@@ -71,7 +74,63 @@ const createModerationMutation = async ({ dataToSign, contentId, contentType, ur
...
+    const resp = await lastValueFrom(sdk.api.auth.signData(dataToSign));
+    const data = {
+      contentId,
+      contentType,
+      data: dataToSign,
+      signature: btoa(String.fromCharCode.apply(null, resp.data.signature)),
+    };
+
+    const status = await createModeration(url, data);
+    if (status === 409) {
```
kenshyx: consider using a switch statement here

ui/hooks/src/use-moderation.ts
```diff
@@ -56,15 +56,18 @@ const createModerationMutation = async ({ dataToSign, contentId, contentType, ur
...
-      throw new Error('Unable to process your request right now. Please try again later');
+
+    switch (true) {
+      case status === 409:
+        throw new Error(`This content has already been moderated by you`);
+      case status === 403:
+        throw new Error('You are not authorized to perform this operation');
+      case status === 400:
+        throw new Error('Bad request. Please try again later');
+      case status >= 400:
```
kenshyx, line 67: there is no `default` case?

josenriagu: no, I have handled all possible status values from the endpoint

kenshyx: a switch statement should always have a default return..., in this case add something that will flag that will help with cases that are not caught by the exact value comparison

kenshyx (approved):

PR #1464, opened 2021-10-28
refactor/use-errors

ghost:
refactor(ui,sdk): update error handling and logging to rely on react-query

chore(): remove useErrors and useLoginState hooks

docs(): update README

## Description
* Refactored most hook queries and mutations to depend on the onError handler for logging instead of try/catch
* catchError operator used for cases when recovery from error made sense
* ErrorInfoCard component refactored to use directly the error property from useQuery
* Few fixes where logging was non-existent or was implemented but did not work correctly
* Removed now unused use-login-state and use-error-state hooks
* Updated hooks package README

## Checklist


ui/hooks/src/use-moderation.ts
```diff
@@ -45,34 +45,28 @@ export type UseModerationParam = {
...
+  const status = await createModeration(url, data);
+
+  switch (true) {
+    case status === 409:
+      throw new Error(`This content has already been moderated by you`);
+    case status === 403:
+      throw new Error('You are not authorized to perform this operation');
+    case status === 400:
+      throw new Error('Bad request. Please try again later');
+    case status >= 400:
```
kenshyx, line 63: you can remove both checks for ( === or > 400) and replace it with default behaviour

ui/hooks/src/use-moderation.ts
```diff
@@ -99,40 +93,37 @@ function useModeration() {
...
+  const status = await createModeration(url, data);
+
+  switch (true) {
+    case status === 409:
+      throw new Error(`This content has already been reported by you`);
+    case status === 403:
+      throw new Error('You are not authorized to perform this operation');
+    case status === 400:
+      throw new Error('Bad request. Please try again later');
+    case status >= 400:
```
kenshyx, line 118: same as my previous comment

ui/hooks/src/use-moderation.ts
ghost: I removed the === 400 case, but I'm not sure 'throw Error' should be the default behaviour :D

kenshyx: hmm, yes, I didn't see that there is no success case

kenshyx: then the `return status` part should be in default case

kenshyx (approved):

PR #1477, opened 2021-11-05
Fix/follow unfollow

ghost:
* Fix bugs relating to "following" functionality caused by query deduplication in useIsFollowingMultiple
* Fix skeleton screen missing in trending widget while data is loading

## Checklist


ui/hooks/src/use-follow.ts
```diff
@@ -49,27 +59,46 @@ export function useIsFollowingMultiple(
   followingEthAddressArray: string[],
 ) {
   const queryClient = useQueryClient();
-  return useQuery(
+  const [refetchPending, setRefetchPending] = React.useState(false);
+  const filteredFollowingEthArray = followingEthAddressArray.filter(_ => _); // often we end up with [undefined] as arg
+  const allAddresses: string[] = queryClient.getQueryData([FOLLOWED_PROFILES_KEY, 'all']) || [];
+  const newAllAddressess = deduplicateArray([...allAddresses, ...filteredFollowingEthArray]);
+  const shouldRefetch = objHash(newAllAddressess) !== objHash(allAddresses);
```
kenshyx, line 66: keep in mind that `objHash` is kinda slow for large data sets

kenshyx (approved):

ui/hooks/src/use-follow.ts
ghost: I thought of just doing .join('') but those could potentially get pretty big 🤔 Possibly just join the last n characters of each address, if this becomes a problem?

PR #1487, opened 2021-11-15
feat(ui): update topbar layout dependent on url path for onboarding

ghost:
## Description
* Update topbar to change its layout to focus mode when on sign-in and sign-up URLs

## Checklist


ui/design/src/components/TopBar/index.tsx
```diff
@@ -128,6 +136,11 @@ const Topbar: React.FC<ITopbarProps> = props => {
     }
   }, [otherAreaItems]);
 
+  const shouldRenderOnboarding = React.useMemo(
```
SeverS: You should use react-router!

ghost: @SeverS I can use the <Route> component for conditional rendering in the top bar, but I sort of still need this flag to pass down to the ProfileMenu to avoid rendering moderation menu options 🤔

SeverS: I've reviewed the designs once again and i think the moderation history link should remain in place, so no need for that flag in the Profile Menu.

ui/design/src/components/TopBar/index.tsx
```diff
@@ -434,8 +450,52 @@ const Topbar: React.FC<ITopbarProps> = props => {
...
+      );
+    }
+  };
+
+  return (
+    <Router>
+      <Switch>
+        <Route
+          path="/auth-app/*"
+          render={() => (
```
SeverS: I think you don't need the render prop, you can pass the component as child

ui/design/src/components/TopBar/index.tsx
```diff
@@ -434,8 +450,52 @@ const Topbar: React.FC<ITopbarProps> = props => {
...
-    </TopbarWrapper>
+      );
+    }
+  };
+
+  return (
+    <Router>
+      <Switch>
+        <Route
+          path="/auth-app/*"
```
SeverS: can you add some tests covering this special case? You can use .skip for now until the code is merged into the next branch but at least we should have them written

SeverS (approved):

PR #1505, opened 2021-12-07
fix(ui): prevent already signed-up user from signing up again

ghost:
## Description
* prevent already signed up user from signin up again
* show message and link to sign-in on step four

## Checklist


sdk/src/auth/index.ts
```diff
@@ -133,12 +133,11 @@ export default class AWF_Auth implements AWF_IAuth {
...
     );
     return check.pipe(
-      tap(response => {
-        if (!response.data || response.data?.errors) {
-          const err = new Error('This ethereum key is not registered');
-          err.name = 'UserNotRegistered';
-          throw err;
-        }
+      tap(() => {
+        this._globalChannel.next({
```
kenshyx: if this replaces the error throw case then it should still keep the guard rule

ghost: I can add it, but from what ive seen it never threw here, getProfile has its own throw which always happened before this

kenshyx: yes, but this is part of the `checkIfSignedUp` method here

kenshyx: I still don't understand the use-case for the event emitted here

kenshyx: this event will be emitted on every `checkIfSignedUp` method call

ghost: Hmmm good point. I used it to keep displaying the loader for connect address, until we know whether to stop the sign-up or not (already registered). But I guess this is something we can live without.

kenshyx: you can still emit it but it needs to be aware of the response, for example you could also attach some data to the event like `isRegistered`

ghost: Turns out there was no actual need for it in the end 🤦‍♂️🤦‍♂️ I'll let myself out

PR #1506, opened 2021-12-09
Fix: v2 signup 1

josenriagu:
## Description

- checks saved token when signup is resumed
- fixes logic for auto network switch on metaMask only
- refactor useConnectProvider hook

## Checklist

### DIFF
--- a/ui/hooks/src/use-injected-provider.ts
+++ b/ui/hooks/src/use-injected-provider.ts
@@ -21,9 +21,9 @@ const getInjectedProvider = async () => {
   return { name: provider.data, details };
 };
 
-const connectProvider = async (provider: EthProviders) => {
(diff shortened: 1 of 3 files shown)


ui/hooks/src/use-injected-provider.ts
```diff
@@ -23,7 +23,7 @@ const getInjectedProvider = async () => {
 
 const connectProvider = async (provider: EthProviders) => {
   const sdk = getSDK();
-  await sdk.services.common.web3.connect(provider);
+  return await sdk.services.common.web3.connect(provider);
```
kenshyx: the await is redundant here, can be removed

kenshyx (approved):

PR #1524, opened 2022-01-11
remove isMobile flag from app-loader and apps

SeverS:
## Checklist


ui/design/src/components/Editor/index.tsx
```diff
@@ -172,11 +172,12 @@ const EditorBox: React.FC<IEditorBox> = React.forwardRef((props, ref) => {
   /**
    * initialise editor with all the required plugins
    */
-  const editor = useMemo(
-    () => withLinks(withTags(withMentions(withImages(withHistory(withReact(createEditor())))))),
-    [],
+  const editorRef = useRef(
+    withLinks(withTags(withMentions(withImages(withHistory(withReact(createEditor())))))),
```
quininez, line 180: why switch to useRef here?

ui/design/src/components/Editor/index.tsx
```diff
@@ -212,7 +213,7 @@ const EditorBox: React.FC<IEditorBox> = React.forwardRef((props, ref) => {
    * set the selection at the end of the content when component is mounted
    */
   useEffect(() => {
-    Transforms.move(editor);
+    Transforms.move(editorRef.current);
```
quininez, line 220: does this change anything? since editor is already assigned to editorRef.current it should be the same thing, no?

ui/design/src/components/Editor/index.tsx
```diff
@@ -367,87 +368,86 @@ const EditorBox: React.FC<IEditorBox> = React.forwardRef((props, ref) => {
...
-    }
-  };
-
   const onKeyDown = useCallback(
     event => {
+      /**
+       * key handler for the mention popover
+       * inserts the mention on tab, enter or space keypress
+       */
+      const selectMention = (event: KeyboardEvent, mentionRange: Range) => {
```
quininez, line 381: what is the advantage of having the function declared here? was it because of codefactor complaining?

ui/design/src/components/Editor/index.tsx
SeverS: To signal that the editor instance does not change between renders

ui/design/src/components/Editor/index.tsx
SeverS: yes, because those methods are used inside the hook only, otherwise they should be listed as dependencies

ui/design/src/components/Editor/index.tsx
quininez: wasn't useMemo providing the same utility?

ui/design/src/components/Editor/index.tsx
SeverS: it's not the same thing! the reference is different when comparing refs and ref.current!

ui/design/src/components/Editor/index.tsx
SeverS: only when using useMemo with empty deps, but apparently codefactor does not like it

ui/design/src/components/Editor/index.tsx
quininez: I was refering to this line https://github.com/AKASHAorg/akasha-world-framework/pull/1524/files#diff-630fe981e4c6d0089d0a1a77d977991138f23c824be3c6f3857f93b1f30c8cb6R183 above, so editor = editorRef.current

SeverS: editorRef.current can change between renders, you can modify it `editorRef.current = new Editor...`. That's why the hook is expecting you to use the actual ref!

quininez: ok, thanks for the explanation

quininez (approved):

PR #1535, opened 2022-01-19
Refactor sdk, graphql api for IC registry

kenshyx:
## Checklist


api/hub/src/resolvers/query.ts
```diff
@@ -227,6 +236,77 @@ const query = {
...
+        });
+        continue;
+      }
+      const integrationInfo = {
+        id: integrationID,
+        name: data.integrationName,
+        author: data.author,
+        latestReleaseId: data.latestReleaseId,
+        integrationType: data.integrationType,
+        enabled: data.enabled,
```
quininez, line 262: this is if we enable users to install the integration?

kenshyx: no, this is if the integration was disabled/enabled from the IC

api/hub/src/resolvers/query.ts
```diff
@@ -227,6 +236,77 @@ const query = {
...
+      };
+      results.push(integrationInfo);
+      await registryCache.set(cacheKey, integrationInfo, 3600);
+    }
+    return results;
+  },
+  getLatestRelease: async (_source, { integrationIDs }, { dataSources }) => {
+    const results = [];
+    const queryPrefix = '0xfi-';
```
quininez, line 271: why is this queryPrefix different from the one in getIntegrationInfo (0xf- vs 0xfi-)?

kenshyx: because we need different keys for the same id

sdk/src/contracts/scripts/deploy-integration-registry.js
```diff
@@ -1,16 +1,30 @@
 /* eslint-disable */
 const { ethers, upgrades } = require('hardhat');
+//const pkgInfos = require('../../../../build/integrations_bucket.json');
 async function main() {
-  //const IntegrationRegistry = await ethers.getContractFactory('IntegrationRegistry');
+  const IntegrationRegistry = await ethers.getContractFactory('IntegrationRegistry');
   //const IntRegistrar = await upgrades.deployProxy(IntegrationRegistry);
```
quininez, line 6: this code is no longer used?

kenshyx: it is used, but I'm keeping it commented to prevent unwanted redeployments until we have an automated process

quininez (dismissed):

SeverS (approved):

PR #1555, opened 2022-02-02
Integration center

quininez:
## Description

- integration center layout
- explore and my apps pages
- integration info page
- integration registry hooks
- app settings hooks

## Checklist


ui/apps/app-center/src/components/app-routes.tsx
```diff
@@ -0,0 +1,96 @@
...
+import routes, { rootRoute, EXPLORE, MY_APPS, APPS, WIDGETS, INFO } from '../routes';
+
+const { Box, Text, BasicCardBox } = DS;
+
+const AppRoutes: React.FC<RootComponentProps> = props => {
+  const { t } = useTranslation();
+
+  return (
+    <Router>
+      <Route path={[routes[EXPLORE], routes[MY_APPS], routes[APPS], routes[WIDGETS]]}>
```
SeverS: I don't understand why the path is an array, what is suppose to do?

quininez: it matches multiple paths to a component, in this case all these routes share the same card element with tabs

SeverS: I couldn't find any reference for path being an array. You can remove this `<Route>` entirely or use `path="*"` if react-router complains about children not being a `<Route>` element

quininez: there is in docs https://v5.reactrouter.com/core/api/Route/path-string-string , and I want to enforce this only on specific routes because on the info page for example we don't have the same component

SeverS:
I have a few questions:
  - Are you sure that the explore page contains those 2 sections `world apps` and `installed apps`? I think that should be `my-apps` page... no?
  - the apps page shouldn't list all integrations in the registry?

quininez:
Yes, the my-apps page contains the 2 sections world apps and installed apps, thats how it is implemented. For the apps page we didn't have designs yet, neither for the widgets, but I can add that to not let it empty

ui/apps/app-center/src/components/app-routes.tsx
SeverS: My concern is that it will be harder and harder to upgrade to v6.
so basically you have 2 routes there:
 - `${routes[INFO]}/:integrationId`
 - wildcard route for everything else
 You can use `<Switch>` element like this
```
<Switch>
  <Route path={`${routes[INFO]}/:integrationId`} />
  <Route path="*">
      <Redirect from="rootRoute" to="/explore" />
      Info Cards Components Here
  </Route>
</Switch>
```

This way, when upgrading to v6 we only replace `<Switch>` with `<Routes>`

SeverS:
ah, ok, my mistake then

josenriagu:
ran the branch locally, looks good ✅

ui/apps/app-center/src/components/app-routes.tsx
ghost: There is an upcoming backwards compatibility layer release for react router v6, so it's likely we won't even need to replace Switch with Routes, but I don't know the exact details

SeverS (approved):

PR #1558, opened 2022-02-09
Feat: Settings app updates

josenriagu:
## Description

- update settings app UI

## Checklist


ui/apps/settings-app/src/interfaces/index.ts
```diff
@@ -0,0 +1,48 @@
+export type SettingsOption =
```
quininez: keep interfaces in the files they are used, unless they are used also by other parts, in that case move them to ui-awf-typings package

quininez (changes requested):

quininez (approved):

PR #1559, opened 2022-02-11
feat: create translation middleware for app-loader

ghost:
## Description

* create translation middleware for app-loader, which can be used to load translation resource path dynamically

* move i18n initialization and configs from apps to app-loader, then pass it down to apps/widgets/extensions, which will now only register their namespaces in their bootstrap method, instead of initializing their own instance

* create nx executor which handles translation extraction, removes the need for individual i18next parser configs and npm scripts, while also allowing for nx computational caching. The executors are **compiled during the bootstrap process**, as subsequent build processes may depend on them being already compiled.

* fix remaining incorrect template strings passed to t method of i18n and replace with interpolation. Update the new nx executor to fail the extraction job if it identifies any incorrect strings passed to the parser.

## Checklist


tools/executors/__tests__/i18n.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const expectedParserConfig = {
+  useKeysAsDefaultValue: true,
+  locales: ['en'],
+  verbose: true,
+  keySeparator: false,
+  nsSeparator: null,
+  defaultNamespace: 'some-namespace',
+  output:
+    '/home/blkns/akasha/akasha-world-framework/tools/executors/locales/$LOCALE/$NAMESPACE.json',
```
kenshyx: this path is absolute, can we use a relative one?

PR #1572, opened 2022-03-03
Refactor/app loader

SeverS:
## Description
Refactor app-loader to using rxjs

Cleanup unused/duplicate props from `register()` function:
    - name
    - sdkModules
    - title

## Breaking changes
 - getMenuItems now returns an observable. The idea is to propagate the changes to menu items without having to recall it every time it changes.
 - Layout widget's `register()` method now returns standard `IWidgetConfig` interface object:
     Previous:
     ```js
      register() {
        return {
           title: '',
           mountsIn: '',
           pluginSlotId: '',
           topbarSlotId: '',
           // etc
        }
      }
     ```
   Now:
   ```js
     register() {
       return {
           title: '',
           mountsIn: '',
           // note the extensions object which is a standard way to
           // declare the extensions that an integration will mount.
           extensions: {
              pluginSlotId: '',
              topbarSlotId: '',
              // etc
           }
       }
    }
   ```

## Integrations that need re-publishing:
 All of them, because of the cleanup and breaking change.

## Testing
 - to be added after

## Checklist


sdk/src/settings/apps.ts
```diff
@@ -75,7 +75,7 @@ class AppSettings implements IAppSettings {
     }
     if (!release?.data?.enabled) {
       this._log.warn(`${app.name} cannot be installed.`);
-      return false;
+      // return false;
```
kenshyx: this needs to be uncommented

kenshyx (approved):

PR #1585, opened 2022-03-07
feat(): implement routing app

ghost:
## Description

* Create a new plugin to store other and expose other app routes, active apps, and menu items
* Update configs in apps to require menuItems field populated
* remove getMenuItems logic
* move navigateTo method into the routing app and refactor

## Checklist


ui/app-loader/src/integrations.ts
```diff
@@ -363,9 +339,20 @@ export const handleExtPointMountOfApps = (
           navigateTo: navigateTo(props.integrationConfigs, logger),
           navigateToModal: navigateToModal,
           getAppRoutes: appName => props.integrationConfigs.get(appName).routes,
-          getMenuItems: () => state$.pipe(getStateSlice('menuItems')),
         },
       });
+
+      // each app must have menuItems exposed in config, widgets do not
+      if (config.menuItems) {
+        dispatchEvent(
```
SeverS: you can use uiEvents instead of custom event.
This way is also easier to keep track of the events

ui/apps/akasha/src/index.tsx
```diff
@@ -24,6 +26,15 @@ export const register: (opts: IntegrationRegistrationOptions) => IAppConfig = op
   title: 'Ethereum World',
   logo: { type: LogoTypeSource.ICON, value: 'appAkasha' },
   i18nNamespace: ['app-akasha-integration', 'ui-lib-feed'],
+  menuItems: {
+    route: rootRoute,
+    label: 'Social',
+    name: 'app-akasha-integration',
```
SeverS: This name should be equal to the name in the registry. Let's avoid redefining names.
In the app-loader when you fire that event, you can specify the app name from there.

ui/apps/routing/src/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+  active?: {
+    [key: string]: Record<string, unknown>;
+  };
+  byArea?: { [key in MenuItemAreaType]?: Array<Record<string, unknown>> };
+}
+
+export class RoutingPlugin {
+  static readonly routeRepository: RouteRepository = {
+    all: {},
+    active: {},
```
SeverS: so this property, `active` has 2 keys, `apps` and `widgets` which are lists of active appNames? My first impression was that it contains active menuItems... Can you make it more explicit? Because `all` and `byArea` are menuItems.

ui/apps/routing/src/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+      }
+    });
+  }
+}
+
+export const register = async () => {
+  RoutingPlugin.initRouteObservation();
+  return {
+    loadingFn: () => Promise.resolve(),
+    name: 'app-routing',
```
SeverS: rm this `name`

ui/apps/routing/src/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+  };
+};
+
+export const getPlugin = async () => {
+  RoutingPlugin.initRouteObservation();
```
SeverS: I think initializing here is enough, no need to do it again in `register()`. I will make register optional so we can remove it entirely

ui/apps/akasha/src/index.tsx
```diff
@@ -24,6 +26,15 @@ export const register: (opts: IntegrationRegistrationOptions) => IAppConfig = op
   title: 'Ethereum World',
   logo: { type: LogoTypeSource.ICON, value: 'appAkasha' },
   i18nNamespace: ['app-akasha-integration', 'ui-lib-feed'],
+  menuItems: {
+    route: rootRoute,
+    label: 'Social',
+    name: '@akashaproject/app-akasha-integration',
```
kenshyx: we can import the name from package.json

kenshyx (approved):

PR #1589, opened 2022-03-14
Tracking/matomo

ghost:
## Description

* Add, update, and fix various matomo analytics calls

## Checklist

### DIFF
--- a/ui/hooks/src/use-login.ts
+++ b/ui/hooks/src/use-login.ts
@@ -2,11 +2,17 @@ import * as React from 'react';
 import getSDK from '@akashaproject/awf-sdk';
 import { events } from '@akashaproject/sdk-typings';
 import { EthProviders } from '@akashaproject/sdk-typings/lib/interfaces';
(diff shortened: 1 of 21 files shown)


ui/hooks/src/use-login.ts
```diff
@@ -187,18 +197,60 @@ export function useSignUp(provider: EthProviders, checkRegistered = false) {
...
+        if (err.code === PROVIDER_ERROR_CODES.UserRejected) {
+          analyticsActions.trackEvent({
+            category: AnalyticsCategories.SIGN_UP_WALLET,
+            action: 'Step 1 Declined',
+          });
+        }
+        onError(err);
+      },
+      onSuccess: () => {
+        analyticsActions.trackEvent({
```
SeverS: I see that `analyticsActions` is marked as optional. I think you should add a guard here to check that `analyticsActions` is defined.

SeverS (approved):

PR #1590, opened 2022-03-15
Feat: sidebar integration

josenriagu:
## Description

- implements sidebar on the feed

## Checklist


ui/apps/profile/src/index.tsx
```diff
@@ -19,9 +19,9 @@ export const register: (opts: IntegrationRegistrationOptions) => IAppConfig = op
   i18nNamespace: ['app-profile', 'ui-lib-feed'],
   menuItems: {
     label: 'Profile',
-    area: [MenuItemAreaType.AppArea, MenuItemAreaType.QuickAccessArea],
+    area: [MenuItemAreaType.QuickAccessArea],
     // routes,
-    logo: { type: LogoTypeSource.AVATAR, value: '' },
+    logo: { type: LogoTypeSource.AVATAR, value: 'person' },
```
quininez: why change this value to 'person'?

ui/apps/app-center/src/index.tsx
```diff
@@ -28,7 +28,7 @@ export const register: (opts: IntegrationRegistrationOptions) => IAppConfig = op
     label: 'Integration Center',
     area: [MenuItemAreaType.QuickAccessArea, MenuItemAreaType.AppArea],
     logo: { type: LogoTypeSource.ICON, value: 'appCenter' },
-    route: rootRoute,
+    route: routes[EXPLORE],
```
quininez: why change the route here?

ui/apps/profile/src/index.tsx
josenriagu: when it was shown on the sidebar, I will revert it now

ui/apps/app-center/src/index.tsx
josenriagu: currently there is nothing showing on the root route as the default page on IC is the explore page. so it makes sense to have same as the route the sidebar will navigate to

quininez: keep the exposed route the rootRoute, the app itself will handle internal redirects (in this case redirecting to the explore page). In case you need the app's root route to be used differently it's good to expose it here, and not a subroute

quininez (approved):

PR #1602, opened 2022-03-31
Integration Center Updates

quininez:
## Description

- updates to IC pages
- notification pill for uninstalling an app
- changes to hooks
- styling changes

## Checklist


ui/apps/app-center/src/extensions/install-modal.tsx
```diff
@@ -103,7 +102,7 @@ const IntegrationInstallModal: React.FC<RootExtensionProps> = props => {
       )}
       errorInfoLabel={t('Please check your network connection and try again.')}
       errorSubInfoLabel={t('Thank you!')}
-      errorSubtitleLabel={t('could not be installed in {{worldName}}.', {
+      errorSubtitleLabel={t('{{integrationName}} could not be installed in {{worldName}}.', {
```
SeverS, line 106: you will need to pass this variable to the second param of tFunc, just like it's with `worldName`

ui/apps/app-center/src/extensions/install-modal.tsx
```diff
@@ -84,9 +84,8 @@ const IntegrationInstallModal: React.FC<RootExtensionProps> = props => {
       dismissLabel={t('Dismiss')}
       modalTitleLabel={t('App Install')}
       integrationName={integrationName}
-      installTitleLabel1={t('To add')}
-      installTitleLabel2={t(
-        'to your World we have to do a few things first. This will take less than a minute.',
+      installTitleLabel={t(
+        'To add {{integrationName}} to your World we have to do a few things first. This will take less than a minute.',
```
SeverS, line 88: `integrationName` should be passed to the second param of tFunc.
`t('{{variableName}}', { variableName: 'my var name' })`

quininez: ok, I changed these now

ui/apps/app-center/src/extensions/integrations-widget.tsx
```diff
@@ -19,38 +20,49 @@ const { Box, ICWidgetCard, ErrorLoader } = DS;
...
   // select user installed apps from list of installed apps
-  const filteredInstalledApps = installedIntegrationsInfoReq.data?.getIntegrationInfo.filter(
-    app => {
-      if (!defaultAppsNamesNormalized?.some(defaultApp => defaultApp.name === app.name)) {
-        return app;
-      }
-    },
-  );
+  const filteredInstalledApps = integrationsInfoReq.data?.getIntegrationInfo.filter(app => {
+    if (defaultIntegrations?.some(defaultApp => defaultApp !== app.name)) {
```
SeverS: I think this check is redundant because an app can either be in defaultIntegrations or in installedAppReq.
What i would do here is a `partition` thus doing only one iteration over the `integrationsInfoReq.data?.getIntegrationInfo`

something like:
```
const {filteredDefault, filteredInstalled} = integrationsInfoReq.data?.getIntegrationInfo.reduce((acc, app) => {
    if(defaultIntegrations.includes(app.name) {
      acc.filteredDefault.push(app)
    } else {
     // is probably installed but you can also check to be sure
     acc.filteredInstalled.push(app);
    }
    return acc;
}, {filteredDefault: [], filteredInstalled: []});
```

quininez: ok, I can change it to this pattern

ui/hooks/src/use-integration-registry.ts
```diff
@@ -112,8 +112,9 @@ const getAllIntegrationsIds = async (offset?: number) => {
 /**
  * Hook to get all the published integrations ids
  */
-export function useGetAllIntegrationsIds(offset?: number) {
+export function useGetAllIntegrationsIds(enabler?: boolean, offset?: number) {
   return useQuery([INTEGRATIONS_KEY, 'ids'], () => getAllIntegrationsIds(offset), {
+    enabled: !!enabler,
```
SeverS, line 117: this should be false by default? if i don't pass the enabler (because it's optional) then it will be false. I would keep it default to true.

quininez: hmm I don't understand, there are 2 ! before the enabler to make sure it's boolean, it's not false

SeverS: I'm referring to the case when you would call this hook without specifying the `enabler` like this:
`useGetAllIntegrationsIds()`.
In this case, `enabler` is `undefined` so `!!undefined` = false

quininez: ok, will add a default true

SeverS (approved):

PR #1609, opened 2022-04-06
UI fixes

quininez:
## Checklist


ui/hooks/src/use-integration-registry.ts
```diff
@@ -106,7 +106,14 @@ export function useGetIntegrationReleaseId(
 const getAllIntegrationsIds = async (offset?: number) => {
   const sdk = getSDK();
   const res = await sdk.api.icRegistry.getAllIntegrationsIds(offset);
-  return res.data;
+  let nextIndex = res.data?.nextIndex?._hex;
```
kenshyx: you can use `res.data?.nextIndex?.toNumber()`

ui/hooks/src/use-integration-registry.ts
```diff
@@ -106,7 +106,14 @@ export function useGetIntegrationReleaseId(
...
+  const integrationIds = res.data?.integrationIds;
+  while (nextIndex && nextIndex !== '0x00') {
```
kenshyx: you can remove `nextIndex !== '0x00'` when nextIndex is a number(0 by default)

ui/hooks/src/use-integration-registry.ts
quininez: if I cast it to number can I still pass it as offset to the sdk call?

kenshyx: you can keep it as it is now `const nextRes = await sdk.api.icRegistry.getAllIntegrationsIds(res.data.nextIndex)`

kenshyx: it uses the response data, not this value from `nextIndex`

quininez: yes, only that is a bug, it should use the nextIndex

kenshyx: it should work with the value casted to number

quininez: but seems it's ok even with the nextIndex passed as a number instead of hex

kenshyx (approved):

PR #1621, opened 2022-04-13
chore(): extract widgets and other fixes

SeverS:
## Description

 - extract my-apps widget and mini-profile widget
 - small other fixes

## Checklist


ui/widgets/mini-profile/src/index.ts
```diff
@@ -0,0 +1,18 @@
...
+} from '@akashaproject/ui-awf-typings/lib/app-loader';
+
+/**
+ * All widgets must export an object like this:
+ */
+export const register: (opts: IntegrationRegistrationOptions) => IWidgetConfig = opts => ({
+  mountsIn: opts.layoutConfig?.widgetSlotId,
+  loadingFn: () => import('./MiniProfileWidget'),
+  activeWhen: (location, pathToActiveWhen) =>
+    pathToActiveWhen('/social-app/post/', false)(location),
```
quininez, line 14: is there a way to get this path from the routing plugin here?

ui/widgets/my-apps/src/index.ts
```diff
@@ -0,0 +1,18 @@
...
+} from '@akashaproject/ui-awf-typings/lib/app-loader';
+
+/**
+ * All widgets must export an object like this:
+ */
+export const register: (opts: IntegrationRegistrationOptions) => IWidgetConfig = opts => ({
+  mountsIn: opts.layoutConfig?.widgetSlotId,
+  loadingFn: () => import('./MyAppsWidget'),
+  activeWhen: (location, pathToActiveWhen) =>
+    pathToActiveWhen('/integration-center/', false)(location),
```
quininez, line 14: also here if it's possible

ui/widgets/mini-profile/src/index.ts
SeverS: hmm, let me check

SeverS: looks like i can't trick it to make it work, the moment the activeWhen is called, it can happen that not all integrations are stored in the routing plugin...

SeverS: so it will return `false` and then no more widget until you navigate away and come back

quininez (approved):

PR #1675, opened 2022-06-01
Refactor: report analytics

josenriagu:
## Description

- move report analytics logic to hook, records event only on success.

## Checklist

### DIFF
--- a/ui/apps/moderation/src/extensions/report-modal.tsx
+++ b/ui/apps/moderation/src/extensions/report-modal.tsx
@@ -10,18 +10,14 @@ import {
   useReasons,
   useGetLogin,
   useReport,
-  useAnalytics,
   ThemeWrapper,
 } from '@akashaorg/ui-awf-hooks';
 import { BASE_REPORT_URL } from '../services/constants';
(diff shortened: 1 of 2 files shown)


ui/apps/moderation/src/extensions/report-modal.tsx
```diff
@@ -49,10 +45,6 @@ const ReportModalComponent = (props: RootExtensionProps) => {
 
   const onReport = React.useCallback(
     (dataToSign: Record<string, string>) => {
-      analyticsActions.trackEvent({
```
SeverS, line 52: Why did you moved the analytics into the hook?

josenriagu: So it tracks, only if the report action was successfully executed. On the modal method, it will send the action regardless of if the mutation was successful or not

SeverS: 1. If is successful then you will see the entry in the moderation app so this event becomes redundant.
2. This event should be recorded regardless of the server's response... (aka The reported is willing to do this action. If there is an error, that's another story)
3. If you really want to do it on success, you can do it in an effect hook without mixing things out.

PR #1687, opened 2022-06-20
Feat: message app DS

josenriagu:
## Description

-  adds design system components for Message App

## Checklist


ui/design/src/components/Editor/index.tsx
```diff
@@ -569,9 +569,11 @@ const EditorBox: React.FC<IEditorBox> = React.forwardRef((props, ref) => {
         className="scrollBox"
         height={minHeight ? { min: minHeight } : undefined}
       >
-        <Box flex={{ shrink: 0 }}>
-          <Avatar src={avatar} ethAddress={ethAddress} margin={{ top: '0.5rem' }} />
-        </Box>
+        {avatar && (
```
quininez: should be a new flag like showAvatar, with default true. Like this you will hide the avatar also in the editor modal if the user doesn't have an avatar set

ui/design/src/components/MessageAppBubbleCard/index.tsx
```diff
@@ -0,0 +1,80 @@
+import * as React from 'react';
```
quininez: the naming of the component should be more generic, and not related to where it will be used, maybe just BubbleCard is enough?

ui/design/src/components/MessageAppBubbleCard/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const time = chatTimestamp?.split('T')[1].split(':');
+
+  return (
+    <Box>
+      <Box direction="row" justify="between" margin={{ bottom: 'xsmall' }}>
+        <Text>{isLoggedUser ? youLabel : sender}</Text>
+        <Text color="secondaryText">{formatRelativeTime(chatTimestamp, locale)}</Text>
+      </Box>
+      <MainAreaCardBox
```
quininez: why use this component here? as the name suggests, this should be used as a container, not for individual smaller components. Why not just use a simple styled box?

ui/design/src/components/MessageAppConvoBody/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+            margin={{ vertical: 'small' }}
+            alignSelf={chat.ethAddress !== loggedUserEthAddress ? 'end' : 'start'}
+            style={{ minHeight: 'min-content', flexShrink: 0 }}
+          >
+            {React.cloneElement(itemCard, {
+              sender: chat.name,
+              status: chat.status,
+              isLoggedUser: chat.ethAddress === loggedUserEthAddress,
+              chatTimeStamp: chat.timestamp,
+              content: (
```
quininez: why is the readOnly Editor here? the itemCard that you are passing should handle the rendering of the content, here you should only pass the content data

quininez (changes requested):

quininez (approved):

PR #1738, opened 2022-09-22
Fix: IC initial state

josenriagu:
## Description

- fixes loading state logic on IC my-apps widget and explore page

## Checklist


quininez (changes requested):
there is still a state where the integration center doesn't load app data properly:
1. go on feed page
3. refresh
4. navigate to IC -> you should see message app as installable, but you only see the placeholder, only if you navigate to another app and return to IC again it will show the proper state

quininez (approved):

PR #1753, opened 2022-10-11
Feat: Enhance editor

didd:
## Description

- Introduces editor enhancements

## Checklist


ui/apps/akasha/src/components/post-page/post-page.tsx
```diff
@@ -281,6 +198,96 @@ const PostPage: React.FC<IPostPageProps & RootComponentProps> = props => {
     [entryData?.delisted, entryData?.moderated, isReported],
   );
 
+  const entryUi = () => {
```
SeverS: I would memoize this function or even extract it to a separate component

ui/apps/akasha/src/extensions/inline-editor.tsx
```diff
@@ -0,0 +1,341 @@
...
+        profileDataReq.status === 'success' &&
+        (action === 'reply' ? (
+          <Box margin="medium" style={{ position: 'relative' }}>
+            {commentEditorUi}
+          </Box>
+        ) : (
+          commentEditorUi
+        ))}
+
+      {publishComment.isLoading && publishComment.variables.postID === props.extensionData.entryId && (
```
SeverS: Is this the pending publish state of a comment? I'm not sure why it should be here, shouldn't be prepended to the list of comments?

ui/apps/akasha/src/components/post-page/post-page.tsx
didd: I've already changed it to a separate component.  I had that in mind :) I will push the commit in a bit.

ui/apps/akasha/src/extensions/inline-editor.tsx
didd: The publish comment mutation is happening inside inline editor and this line depends on the state of `publishComment`. That's why I brought it here. Another way could be I could publish an event for this and it could stay prepended to the list of comments. What do you think ?

SeverS: You can have the same logic as for the post publishing where useMutationListener hook is used, no need for an extra event. The post publishing also have pending state

didd: That's a very clean approach. Ok, I will use that.

ui/apps/search/src/components/routes/search-page.tsx
```diff
@@ -288,7 +288,10 @@ const SearchPage: React.FC<SearchPageProps> = props => {
       showLoginModal();
       return;
     } else {
-      props.navigateToModal({ name: 'editor-modal', embedEntry: entryId });
+      navigateTo?.({
+        appName: '@akashaorg/app-akasha-integration',
+        getNavigationUrl: () => `post/${entryId}?action=repost`,
```
SeverS: you can use the same approach as in the `ui/lib/feed/src/component/entry-feed.tsx` -> `getNavigationUrl: navRoutes => ${navRoutes.Post}/${entryId}?action=repost`

SeverS (approved):

PR #1760, opened 2022-10-17
feat: add breadcrumb navigation extension

eldiegod:
## Checklist


ui/apps/routing/src/extensions/breadcrumb-navigation.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const route = Object.keys(routes).find(route => url.includes(routes[route]));
+  return route;
+};
+
+const stackLimit = 4;
+const BreadcrumbNavigation: React.FC<RootExtensionProps> = props => {
+  const [stack, setStack] = React.useState([location.href]);
+
+  React.useEffect(() => {
+    window.addEventListener('single-spa:before-routing-event', (evt: CustomEvent) => {
```
kenshyx: This listener supports also setting a signal, you can use `AbortController` to call `.abort()` when you need to remove the listener. This is a better approach so you don't end up cleaning another listener.

ui/apps/routing/src/extensions/breadcrumb-navigation.tsx
```diff
@@ -29,19 +29,25 @@ const BreadcrumbNavigation: React.FC<RootExtensionProps> = props => {
...
 
   React.useEffect(() => {
-    window.addEventListener('single-spa:before-routing-event', (evt: CustomEvent) => {
-      const { newUrl } = evt.detail;
+    const controller = new AbortController();
+    const signal = controller.signal;
+    window.addEventListener(
+      'single-spa:before-routing-event',
+      (evt: CustomEvent) => {
+        const { newUrl } = evt.detail;
```
kenshyx: what if `newUrl` does not exist or `evt.detail`

kenshyx: Also, I think that you should keep the `oldUrl` in the stack not the `newUrl` since you want to navigate back

eldiegod: As I understood from singleSPA that's how the event works and it always has `detail` with `oldUrl` and `newUrl`

the `oldUrl` is saved when the component is mounted because we need it to show something even before the user has navigated somewhere

eldiegod:
Not sure why couldflare is falling tho, can't read the logs either

ui/apps/routing/src/extensions/breadcrumb-navigation.tsx
kenshyx: Also the last segment in the breadcrumb(current route) is just a text, you can't click on it

eldiegod: yeah thats the current url the user is at: https://www.figma.com/file/fay7vbao49vi2LYYNvhQ7E/%5BDesign%5D-Replies?node-id=31%3A49337

ui/apps/routing/src/extensions/breadcrumb-navigation.tsx
```diff
@@ -172,21 +181,27 @@ const BreadcrumbNavigation: React.FC<RootExtensionProps> = props => {
...
           margin={{ bottom: 'xsmall', right: 'xsmall' }}
           width="auto !important"
           align="center"
           direction="row"
         >
-          <StyledBackIcon type="arrowLeft" size="sm" />
+          {backButtonActive ? (
+            <StyledBackIcon type="arrowLeft" size="sm" />
+          ) : (
+            <Icon type="akasha" size="sm" />
```
quininez: the idea here was to have the icon of the specific app that you are on now, just like in the sidebar where each app has also an icon, not only a fixed akasha logo

eldiegod: ah ok, just added it!

SeverS (dismissed):

PR #1765, opened 2022-10-21
Feat: Implement new replies UI/UX

didd:
## Description

- Create replies page

## Checklist


typings/src/ui/ui-events.ts
```diff
@@ -42,6 +42,7 @@ export type EventDataTypes = {
   name: string;
   version?: string;
   entryId?: string;
+  commentId?: string;
```
SeverS: Can you use the entryId here instead of creating a new key? This is the way it was suppose to work (yeah, I think the naming is misleading). We'll need to rename the entryId to itemId so it should reflect that it's generic. If it's not much you can do this change (or leave it for future but maybe add a comment there instructing to use the entryId)

ui/apps/akasha/src/components/post-page/common/main-entry.tsx
```diff
@@ -88,19 +92,20 @@ export function PostEntry({
...
           />
         );
       case 'edit':
         return (
           <Extension
-            name="inline-editor_postedit"
+            name={`inline-editor_postedit_${entryData?.entryId}`}
             uiEvents={uiEvents}
-            data={{ entryId: entryData.entryId, action: 'edit', isShown: true }}
+            data={{ entryId: postId, commentId, action: 'edit' }}
```
SeverS: Do you need to pass the postId? commentData does have a `parent` property, right?

typings/src/ui/ui-events.ts
didd: For the inline editor, sometimes we need both the post id and comment id and entryId is scalar in this case. There is generic type in the `EventDataTypes` but I needed the type check when I am mounting the inline editor using `Extension` component. May be we do `entryId?: string | Record<"postId" | "commentId", string>` as a type ?

ui/apps/akasha/src/components/post-page/common/main-entry.tsx
didd: Correct. This ties in with what you suggested above. Hmm so I thought, if I pass in the postId and commentId that will be enough to determine if I am having a reply to a post or a reply to a comment and render the appropriate editor. However, with you suggestion above I can simply pass the entry type to handle this then all I need is `entryId` to resolve the comment you shared above.

didd: My thought process above was from the perspective of making the code readable but I agree with your suggestions. I will change the code accordingly.

typings/src/ui/ui-events.ts
didd: Having read your comment below, I will take out the commentId and leave only entryId as it is.

ui/apps/akasha/src/components/post-page/common/main-entry.tsx
SeverS: yep, that's exactly how it suppose to work (except the unfortunate naming).
Question: How is a reply to a post different than a reply to a comment? I feel like there is a big branching happening but I don't know if it's really necessary.

didd: Hmm I think I phrased things a bit vaguely. So in terms of reply to a post and reply to a comment the only difference is that during a reply to a comment we plugin in the 'replyTo' field. Other than that there is no difference at all. 

But in terms of dividing the inline editor into post editor and reply editor there are a number of benefits. 
1. Post editor deals with editing, posting, reposting and replying to a post
2. Reply editor deals with editing and replying to a comment

In both those cases the api's we are dealing with are different and it makes the code easier to reason about by having this two kind of editors. Other than that we have a base component where both this editors share consisting of common functionalities.

SeverS: Alright, just make sure to specify somewhere that the post editor is also used to reply to a post, my initial thought was that is only used for posts and the reply editor for all things that are `replies`

SeverS (changes requested):
Can be merged after the above changes

didd:
> fix: address PR review

@SeverS On the latest commit, I also fixed two more issues that I noticed. 
1. The 'Save' to bookmark button was hidden on a post page
2. During loading of feed page the loading placeholder were crammed up and the `itemSpacing` wasn't applied 

In addition, I moved `ReplyError` component from reply editor into base. Reply error should be shown when replying to a post and comment. 

Further I noticed one more issue that isn't related to this PR but an observation I made which we can discuss on the dev channel. During posting, the pending post UI stays a bit longer after the post appears on the feed. This happens on a slower connection.

SeverS (approved):

PR #1790, opened 2022-11-15
refactor: ui-events type names

eldiegod:
Issue 694

## Checklist


typings/src/ui/ui-events.ts
```diff
@@ -33,17 +33,29 @@ export enum EventTypes {
...
+/**
+  To be used as per example `EntityTypesMap[EntityTypes.ENTRY]` to get the name of the entity type
+ */
+export const EntityTypesMap = {
+  [EntityTypes.ENTRY]: 'entry',
+  [EntityTypes.PROFILE]: 'profile',
+  [EntityTypes.COMMENT]: 'comment',
+  [EntityTypes.TAG]: 'tag',
+  [EntityTypes.ARTICLE]: 'article',
+} as const;
```
josenriagu: having Entry as entry here and comment as comment conflicts with what you have in index.ts lines 79-87. If an example, maybe we could use the correct values too?

typings/src/ui/ui-events.ts
```diff
@@ -33,17 +33,29 @@ export enum EventTypes {
 
 export const enum EntityTypes {
   ENTRY = 0,
```
josenriagu: Can we eliminate entirely the need for Entry and just call is POST?
Also change COMMENT to REPLY? We can update this EntityTypes as we define more in the app

josenriagu (changes requested):

typings/src/ui/index.ts
```diff
@@ -73,3 +75,13 @@ export enum LEGAL_DOCS {
...
+
+export const ModerationEntityTypesMap = {
+  [EntityTypes.PROFILE]: 'account',
+  [EntityTypes.ENTRY]: 'post',
+  [EntityTypes.COMMENT]: 'reply',
+
+  // @TODO: add support for tag type, when tag moderation is implemented
+  [EntityTypes.ARTICLE]: 'article',
+  [EntityTypes.TAG]: 'tag',
+} as const;
```
josenriagu: You may move all Moderation app related typings to moderation.ts

ui/apps/akasha/src/components/__tests__/reply-page.spec.tsx
```diff
@@ -23,8 +23,8 @@ const MockedInlineEditor = ({ action }) => (
     {...genAppProps()}
     extensionData={{
       name: 'name',
-      entryId: '01gf',
-      entryType: EntityTypes.COMMENT,
+      itemId: '01gf',
+      itemType: EntityTypes.COMMENT,
```
josenriagu: EntityTypes.REPLY

josenriagu (approved):

PR #1793, opened 2022-11-16
Test: add unit test coverage to feed and post pages

didd:
## Description

- Add unit test coverage to feed and post pages

## Checklist

- [X] I have added tests to cover my changes


ui/apps/akasha/src/components/__tests__/feed-page.spec.tsx
```diff
@@ -51,4 +53,32 @@ describe('< FeedPage /> component', () => {
     expect(screen.getAllByTestId('avatar-image')).not.toBeNull();
     expect(screen.getByText(/Share your thoughts/i)).toBeInTheDocument();
   });
+
+  it('should show draft post from local storage and clear it', async () => {
+    const loginState = genLoggedInState(true);
+    localStorage.setItem(
```
kenshyx: I think we should mock localStorage(maybe with Map object), we don't need to test if the localStorage is saving the data properly. It also saves us from having to do cleanup after running the tests(clear item from localStorage)

didd: Hmm actually I am not testing if the localStorage is saving the data properly. This are two things, one is I am testing the feed page showing the saved draft post. And then testing clearing of this saved draft post.

To avoid this confusion, I've split the test case into two.

didd: I've mocked the localStorage based on what we discussed on the review call.

ui/apps/akasha/jest.setup.js
```diff
@@ -88,3 +94,5 @@ mockIntersectionObserver.mockReturnValue({
 });
 
 global.IntersectionObserver = mockIntersectionObserver;
+
+Object.defineProperty(window, 'localStorage', { value: localStorageMock });
```
kenshyx: We should not pollute the global scope, no need to overwrite it, just import the mocked localstorage where you need it in the tests

didd: Hmm fair enough, I have moved that line into the feed spec. But the way to mock local storage is to just use: 
`Object.defineProperty(window, 'localStorage', { value: localStorageMock });`

And then that will be applied only in the context of the feed spec.

didd: localStorage object can't be directly spied on as long as I am aware. The best you can do is something like
`jest.spyOn(Storage.prototype, 'setItem');` That is spying on each method one by one but I don't see a reason to do that. Instead of complicating the code `Object.defineProperty(window, 'localStorage', { value: localStorageMock });` does what needs to be done safely.

didd: :) but if the name 'window' is disturbing. I've just changed it to global.

ui/apps/akasha/src/extensions/inline-editor/post-editor/index.tsx
```diff
@@ -159,10 +159,10 @@ export function PostEditor({ appName, postId, pubKey, singleSpa, action, draftSt
       setEditorState={(value: IEntryData['slateContent']) => {
         if (!canSaveDraft) return;
         if (isEqual(value, editorDefaultValue)) {
-          draft.clear({ pubKey, appName });
+          draft.clear();
           return;
         }
-        draft.save({ pubKey, appName, content: value });
+        draft.save();
```
kenshyx: you still need to pass the content to be saved here

didd: Ouch! I just changed the constructor function into a class to get solid typing.

ui/apps/akasha/src/extensions/inline-editor/utils/draft.ts
```diff
@@ -0,0 +1,48 @@
...
+interface IDraft {
+  storage: IDraftStorage;
+  appName: string;
+  pubKey: string;
+  save(content: IEntryData['slateContent']): void;
+  get(): void;
+  clear(): void;
+}
+
+const getDraftKey = (appName: string, pubKey: string) => `${appName}-${pubKey}-draft-item`;
```
kenshyx: this can be a static method on the Draft class

kenshyx (dismissed):

kenshyx (approved):

PR #1796, opened 2022-11-17
close buttons for the merge +  private alpha notifications

themonster2015:
## Description
* Add icon buttons to allow users to close the notification messages they get about the merge and our private alpha.

## Checklist


ui/hooks/src/use-close-notification.ts
```diff
@@ -0,0 +1,34 @@
...
+          return initialValue;
+        }
+      } catch (err) {
+        // console.log(err);
+        return initialValue;
+      }
+    }
+  });
+
+  const setValue = value => {
```
kenshyx: can we add a type to the value parameter?

ui/widgets/layout/src/components/layout-widget.tsx
```diff
@@ -196,20 +201,30 @@ const Layout: React.FC<RootComponentProps> = props => {
...
+                            'But some functionalities such as claiming ENS names won’t be possible',
+                          )}. ${t('We are working hard on mitigating any issues')}. ${t(
+                            'Bear with us 🙏🏽',
+                          )}.`}
+                        </Text>
+                      </Box>
+                      <Icon
+                        type="close"
+                        clickable={true}
+                        onClick={() => setCloseNotification(true)}
```
kenshyx: this method should be extracted, otherwise you're going to create a new function on every rerender

ui/widgets/layout/src/components/layout-widget.tsx
```diff
@@ -23,9 +23,10 @@ import {
   FocusedPluginSlot,
   CookieWidgetSlot,
 } from './styled-slots';
-import { usePlaformHealthCheck } from '@akashaorg/ui-awf-hooks';
+import { usePlaformHealthCheck, useCloseNotification } from '@akashaorg/ui-awf-hooks';
 
 const { Box, BasicCardBox, Icon, styled, Text, Extension } = DS;
+const CLOSE_THE_MERGE_NOTIFICATION_FLAG = 'close-the-merge-notification';
```
kenshyx: there should be a mapping of the registered flags when these are passed to the hook, otherwise there could be multiple components registering the same flag name

ui/hooks/src/use-close-notification.ts
```diff
@@ -0,0 +1,34 @@
...
+    try {
+      // Allow value to be a function so we have same API as useState
+      const newClosedValue = value instanceof Function ? value(closed) : value;
```
SeverS: If I understand correctly you want to support callback based state setter here.
The correct approach would be to use the actual previous state from the setter function:
```
setClosed(prev => {
  const result = typeof value === 'function' ? value(prev): value;
  // set the localstorage..
  return result 
});
```
But i don't feel like we need this functionality... I would prefer to have simple logic instead, so keep it simple, don't replicate the apis, provide good typings and that's it :)

ui/hooks/src/use-close-notification.ts
```diff
@@ -0,0 +1,34 @@
+import { useState } from 'react';
+
+export function useCloseNotification(key: string, initialValue = false) {
+  const [closed, setClosed] = useState(() => {
```
SeverS: I would prefer to extract this callback function outside of the hook and provide the initialValue as param

ui/hooks/src/use-close-notification.ts
SeverS: The value here should be `React.Dispatch<React.SetStateAction<boolean>>` which is not something i'm expecting to see in a simple setter function. Better stick to just `boolean` and remove extra logic that is not needed at the moment

ui/hooks/src/use-close-notification.ts
themonster2015: Thanks Sever. I have updated the logic.

ui/hooks/src/use-close-notification.ts
themonster2015: I have updated the function accordingly.

ui/widgets/layout/src/components/layout-widget.tsx
themonster2015: Thanks. I have implemented it in a different way :)

kenshyx:
@all-contributors please add @themonster2015  for code

kenshyx:
@all-contributors please add @basmatiii for a11y, design and userTesting

quininez (approved):

PR #1802, opened 2022-11-18
Feat: add ability to scroll top

didd:
## Description

- Add ability to scroll top

## Checklist


ui/design/src/components/EntryList/index.tsx
```diff
@@ -38,15 +44,48 @@ const EntryList = (props: EntryListProps) => {
   const items = (page: IEntryPage) =>
     viewAllEntry ? page?.results.slice(0, viewAllEntry.limit) : page?.results;
 
+  React.useEffect(() => {
+    const observer = new IntersectionObserver(entries => {
+      const entry = entries[0];
+      if (entry.isIntersecting) {
+        scrollTopWrapperRef.current.style.marginLeft = `${
```
kenshyx: hmm, we should not have styling inside hooks ...

didd: Hmm, isn't this a side effect ? And in addition since the scroll to top button is positioned fixed we need to set its margin left dynamically so that it stays within the feed widget container.

kenshyx: Let's discuss about this practice on Monday

SeverS (approved):

PR #1818, opened 2022-11-28
fix  useDismissedCard hook tests

themonster2015:
## Description
* Fix the unit tests for the useDismissedCard hook.

## Checklist


ui/hooks/src/use-dismissed-card.ts
```diff
@@ -1,19 +1,28 @@
...
+
+const writeToLocalStorage = (storage: IStorage, key: string, value: string[]): void => {
   if (storage) {
     storage.setItem(key, JSON.stringify(value));
   }
 };
 
-export function useDismissedCard(): [string[], (newClosedValue: string) => void] {
+export function useDismissedCard(
+  mockStorage?: IStorage,
```
didd:
```suggestion
  storage?: IStorage,
```
The name `mockStorage` makes sense in the context of testing not here :)

didd:
`storage.ts` in `utils` folder is going to be a dangling file which isn't used, I suggest we remove it.

ui/hooks/src/use-dismissed-card.ts
themonster2015: :) I changed it to draftStorage.

themonster2015:
> `storage.ts` in `utils` folder is going to be a dangling file which isn't used, I suggest we remove it.

Yes I missed this one.

ui/hooks/src/use-dismissed-card.ts
didd: I suggest `statusStorage` rather than draftStorage as your intent is saving status.

ui/hooks/src/use-dismissed-card.ts
```diff
@@ -1,19 +1,28 @@
...
+
+const writeToLocalStorage = (storage: IStorage, key: string, value: string[]): void => {
   if (storage) {
     storage.setItem(key, JSON.stringify(value));
   }
 };
 
-export function useDismissedCard(): [string[], (newClosedValue: string) => void] {
+export function useDismissedCard(
+  draftStorage?: IStorage,
```
didd: I suggest `statusStorage` rather than draftStorage as your intent is saving status :)

themonster2015: Yes this makes sense :)

didd (approved):

PR #1844, opened 2023-01-20
Feat: new DS core components

josenriagu:
## Description

- new DS core components

## Checklist


ui/design-system-core/src/interfaces/avatar.interface.ts
```diff
@@ -0,0 +1,22 @@
+export type AvatarSrc = { url?: string; fallbackUrl?: string };
```
quininez: its better if we keep the interface definitions in the file where they are used

ui/design-system-core/src/components/ModerationAppErrorCard/index.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
```
quininez: we should keep more generic components in the design system core, so either make it something like ErrorCard and make it more generic or implement it directly in the app

ui/design-system-core/src/interfaces/basicCardBox.interface.ts
```diff
@@ -0,0 +1,11 @@
+export interface IBasicCardBox {
```
quininez: same thing for interfaces, keep them with the component

ui/design-system-core/src/interfaces/moderationAppErrorCard.ts
```diff
@@ -0,0 +1,14 @@
+export interface IModerationAppErrorCardProps {
```
quininez: interface -> component

ui/design-system-core/src/interfaces/basicpopover.interface.ts
```diff
@@ -0,0 +1,9 @@
+export type OverflowType = 'auto' | 'hidden' | 'scroll' | 'visible';
```
quininez: interface -> component

ui/design-system-core/src/interfaces/appAvatar.interface.ts
```diff
@@ -0,0 +1,6 @@
+import { IntegrationTypes } from '@akashaorg/typings/ui';
```
quininez: interface -> component

quininez (changes requested):

ui/design-system-core/src/interfaces/avatar.interface.ts
josenriagu: Alright. I'll fix those

quininez (approved):

PR #1845, opened 2023-01-20
Feat/button component

themonster2015:
## Checklist


ui/design-system-core/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+const baseStyles = apply`text-center flex items-center justify-center	font-medium md:max-md:w-full hover:drop-shadow-md`;
+const textOnlyStyles = apply`
+${baseStyles}
+text-secondary-light dark:text-secondary-dark hover:opacity-50 dark:hover:text-white
+disabled:opacity-50 disabled:cursor-not-allowed space-x-2
+`;
+
+export const Button: React.FC<PropsWithChildren<ButtonProps>> = props => {
+  const { label, size, greyBg, primary } = props;
+  const loading = props.loading === true;
```
quininez: no need for another variable with the same name for loading, you can use directly the one from props

ui/design-system-core/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const textonly = props.textonly === true || props.textonly !== undefined;
```
quininez: same thing here, just make use of the variable from props

ui/design-system-core/src/components/Button/Button.tsx
themonster2015: thanks, @quininez for pointing it out. I missed it when doing refactoring 😁

quininez (approved):

PR #1853, opened 2023-02-08
chore(): cleanup

SeverS:
## Checklist


ui/hooks/src/use-profile.ts
```diff
@@ -335,7 +335,7 @@ const saveAvatar = async (
       return {
         provider: ProfileProviders.EWA_BASIC,
         property: ProfileProviderProperties.AVATAR,
-        value: res.CID,
+        // value: res.CID,
```
kenshyx: you can use optional chaining here to keep the value prop

kenshyx (approved):

PR #1868, opened 2023-02-26
Feat/pill component

themonster2015:
## Description

Create the Pill and Snackbar components.
## Checklist


ui/design-system-core/src/components/Icon/index.tsx
```diff
@@ -18,6 +18,7 @@ export interface IconProps {
   disabled?: boolean;
   testId?: string;
   styling?: string;
+  onClick?: () => void;
```
didd:
```suggestion
  onClick?: (event: React.SyntheticEvent<Element, Event>) => void;
```

ui/design-system-core/src/components/Snackbar/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export type snackBarType = 'alert' | 'caution' | 'success' | 'info';
+
+export interface ISnackbar {
+  title?: string;
+  description?: string;
+  actionButtonLabel?: string;
+  type: snackBarType;
+  handleButtonClick?: () => void;
+  handleDismiss?: () => void;
```
didd:
```suggestion
  handleDismiss?: (event: React.SyntheticEvent<Element, Event>) => void;
```

ui/design-system-core/src/components/Snackbar/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { getColorLight, getColorDark } from './getColor';
+
+export type snackBarType = 'alert' | 'caution' | 'success' | 'info';
+
+export interface ISnackbar {
+  title?: string;
+  description?: string;
+  actionButtonLabel?: string;
+  type: snackBarType;
+  handleButtonClick?: () => void;
```
didd:
```suggestion
  handleButtonClick?: (event: React.SyntheticEvent<Element, Event>) => void;
```

themonster2015: thanks Didd 👍

quininez (commented):
you have conflicts with next that need to be fixed

ui/design-system-core/src/components/Icon/index.tsx
```diff
@@ -18,6 +18,7 @@ export interface IconProps {
   disabled?: boolean;
   testId?: string;
   styling?: string;
+  onClick?: (event: React.SyntheticEvent<Element, Event>) => void;
```
didd:
```suggestion
```

Let's keep the API of the icon component lean and clean. You can handle the `onClick` event as follows:

```
<button onClick={...}>
    <Icon .../>
</button>
```

Once I finish the work on enhancing the button component I will replace the above suggestion with something like :

```
<Button onClick={...}>
    <Icon .../>
</Button>
```

It's a better API design to wrap every design system component which needs the `onClick` handler with the button element or component. That way we maintain the semantic meaning of each component. Buttons handle click events and the other components address a specific concern. In the case of icons the concern is simply to render icons of different sizes and styles.

jenvonau: Thanks, Didd. I have made the change.

themonster2015: Thanks, Didd. I have made the change.

quininez: But why is it a better design like this? Is there any specific advantage of keeping semantic purity that outweighs the hassle of having to wrap components like divs or icons  whenever you want a simple click handler? @didd

didd: @quininez what makes it better API design is to gain the benefits of better SEO and accessibility as a result of using semantic HTML. As a result of using button we can gain a native support for keyboard navigation, better support for screen readers, we can get a native support for disabled state of an element, better unit testing experience and the like. 

And in this PR's case we have an icon button which we can use but the button component will need to be enhanced first then we can have a choice of using the onClick handler as 
`<Button icon={...} onClick={...} ...` or 
`<Button onClick={...}> <Icon .../> </Button> `

ui/design-system-core/src/components/Snackbar/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+      </span>
+      <div className={tw('w-11/12')}>
+        <Text variant="button-md" color={tw(textcolor)}>
+          {title}
+        </Text>
+        <Text variant="body2" color={tw(textcolor)}>
+          {description}
+        </Text>
+        {actionButtonLabel && (
+          <span onClick={handleButtonClick} className={tw(`cursor-pointer`)}>
```
didd:
```suggestion
          <button onClick={handleButtonClick}>
```

Latter on I will change it to the button component `Button`. You can check [this](https://www.builder.io/blog/buttons) and [this](https://javascript.plainenglish.io/stop-using-divs-for-buttons-87a0b3d7945e) links to get a better insight on why we prefer button and anchor tag over other elements for handling click events.

ui/design-system-core/src/components/Icon/index.tsx
themonster2015: That's a nice explanation Didd 👍

quininez (dismissed):

quininez (approved):

PR #1870, opened 2023-03-01
Feat: enhance profile cards

didd:
## Description

- Enhance profile cards
- Enhance icons, and app icons

## Checklist

### DIFF
--- a/ui/design-system-core/src/components/Icon/index.tsx
+++ b/ui/design-system-core/src/components/Icon/index.tsx
@@ -39,8 +39,10 @@ export const iconTypes: IconType[] = [
   'twitter',
 ];
 
+/*@TODO: change all custom icons into ones in the design system and avoid the following */
 const fillOnlyIcons: IconType[] = ['akasha', 'appModeration'];
 
+/*@TODO: change all custom icons into ones in the design system and avoid the following */
 const fillAndStrokeIcons: IconType[] = ['twitter', 'telegram', 'discord'];
 
 const Icon: React.FC<IconProps> = props => {
@@ -75,8 +77,10 @@ const Icon: React.FC<IconProps> = props => {
     return null;
   }
 
+  /*@TODO: change all custom icons into ones in the design system and avoid the following */
   const isFillOnlyIcon = fillOnlyIcons.includes(type);
 
+  /*@TODO: change all custom icons into ones in the design system and avoid the following */
   const isFillAndStrokeIcon = fillAndStrokeIcons.includes(type);
 
   const plainStyle = plain
--- a/ui/design-system-core/src/components/Icon/index.tsx
+++ b/ui/design-system-core/src/components/Icon/index.tsx
@@ -1,8 +1,11 @@
 import React from 'react';
-import { tw } from '@twind/core';
-
 import * as HeroIcons from '@heroicons/react/24/outline';
 import * as CustomIcons from './akasha-icons';
+import Stack from '../Stack';
+import { apply, tw } from '@twind/core';
+import { ICON_SIZE_MAP, Size } from '../types/common.types';
+import { getWidthClasses } from '../../utils/getWidthClasses';
+import { getHeightClasses } from '../../utils/getHeightClasses';
 
 export type IconType = CustomIcons.CustomIconTypes | keyof typeof HeroIcons;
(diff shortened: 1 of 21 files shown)


ui/widgets/sidebar/src/components/sidebar-new.tsx
```diff
@@ -182,13 +182,7 @@ const Sidebar: React.FC<ISidebarProps> = props => {
...
             <div key={icon.name + idx} className={tw('mr-4')}>
               <a href={icon.link} target="_blank" rel="noreferrer noopener">
-                <div
-                  className={tw(
-                    'w-10 h-10 flex items-center justify-center rounded-full bg-grey9 dark:bg-grey3',
-                  )}
-                >
-                  <Icon type={icon.name} isCustomIcon={true} plain={true} accentColor={true} />
-                </div>
+                <AppIcon placeholderIconType={icon.name} size="sm" plain accentColor />
```
josenriagu, line 185: Can we have both plain and accentColor props set in one icon?

ui/design-system-core/src/components/Icon/index.tsx
```diff
@@ -34,46 +39,91 @@ export const iconTypes: IconType[] = [
...
-  const className = `flex items-center justify-center select-none ${
-    plain ? `[& *]:${svgPrefix}-black dark:[& *]:${svgPrefix}-white` : ''
-  } ${color ? `[& *]:${svgPrefix}-${color}` : ''} ${
-    accentColor ? `[& *]:${svgPrefix}-secondary-dark` : ''
-  } ${clickable && !disabled ? `cursor-pointer hover:[& *]:${svgPrefix}-secondary-dark` : ''}`;
+  const plainStyle = plain
+    ? `${isFillOnlyIcon ? '' : '[&>*]:stroke-black dark:[&>*]:stroke-white'} ${
+        isFillAndStrokeIcon ? '[&>*]:fill-transparent' : ''
+      }`
+    : '';
```
josenriagu, line 90: here plain prop sets to black and white as the case may be.

ui/design-system-core/src/components/Icon/index.tsx
```diff
@@ -34,46 +39,91 @@ export const iconTypes: IconType[] = [
...
+  const colorStyle = color ? `[&>*]:stroke-${color} [&>*]:fill-${color}` : '';
+  const accentColorStyle = accentColor
+    ? `${isFillOnlyIcon ? '' : '[&>*]:stroke-secondary-light dark:[&>*]:stroke-secondary-dark'} ${
+        isFillAndStrokeIcon ? '[&>*]:fill-transparent' : ''
+      }`
+    : '';
```
josenriagu, line 96: and `accentColor` sets the corresponding secondary colors.

josenriagu (changes requested):
Looks good

ui/widgets/sidebar/src/components/sidebar-new.tsx
didd: I am not seeing the need to have this plain prop. The default can be set as plain. That will be the base style. Then I will convert the svg's for all custom icons to the once we have on the design system that will make the Icon component leaner and cleaner. All this will be coming on my next PR.

ui/design-system-core/src/components/Icon/index.tsx
didd: The plain prop won't be necessary as it can be set as a base style which can be overriden by accentColor prop using apply function. I will introduce this change on my next PR.

ui/design-system-core/src/components/Icon/index.tsx
didd: This part will become leaner when we replace the existing custom icons with the once on the design system. Then we won't need to check for things like `isFillOnlyIcon` and `isFillAndStrokeIcon`. I will have this changes on my next PR.

ui/widgets/sidebar/src/components/sidebar-new.tsx
josenriagu: Gotcha, I thought so too, since we may easily pass color to the icons or specify accent color.

josenriagu (approved):

PR #1871, opened 2023-03-01
Feat: moderation app overview page

josenriagu:
## Description

- update moderation app overview page

## Checklist


ui/design-system-core/src/components/Text/index.tsx
```diff
@@ -33,6 +33,7 @@ export type TextProps = {
   truncate?: boolean;
   breakWord?: boolean;
   weight?: Weight;
+  onClick?: () => void;
```
didd: As a component API design, a `Text` component doesn't have event handler as a prop as its semantic purpose is to display simple text. You can check this pattern [here](https://polaris.shopify.com/components/typography/text) and [here](https://mui.com/material-ui/api/typography/) for more reference. 

The proper component for your case is most likely a button component with text variant. This has a semantic meaning as the text element can be clicked so it's a text button. 

I am working on enhancing the button component and in the mean time you can wrap [this](https://github.com/AKASHAorg/akasha-framework/pull/1871/files#diff-528c8c266b88a807b517ced5c25765443a29d8d287b700eeacdf3e1f5f838e66R67-R76) `Text` component with button. Something like `<button onClick={...}> ... </button>` and remove the `onClick` prop from `Text` component. 

Once I am done enhancing the button component I will change that into a button component with text variant.

josenriagu: Alright, I will clean that up now

didd (dismissed):

didd (approved):

PR #1950, opened 2023-06-15
Feat: create install and uninstall app

didd:
## Description

- Create install and uninstall app

## Checklist


ui/design-system-components/src/components/InstallApp/install-icon.tsx
```diff
@@ -0,0 +1,28 @@
...
+      />
```
kenshyx: can we avoid base64 encoded images? :/

kenshyx: also https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href is deprecated

didd: Agreed. I will ask Basma for a WebP version. I took the svg directly from Figma.

kenshyx: webp might not work with svgImage

didd: I had a chat with Basma and she explained those are actually lottie file animation. Never heard about this before :) And this lottie files are available elsewhere in the apps as well. I found out that there is a [react component](https://github.com/LottieFiles/lottie-react) for such files. We will need to discuss the implementation detail and in the mean time I have replaced the svg's with a placeholder.

didd: Finally, I have converted all the icons into WebP and used `img` tag to display them. The animations are complex and creating them in css will not work.

kenshyx (approved):

kenshyx (approved):

PR #1959, opened 2023-06-20
Fix/sign in styling

SeverS:
## Checklist


ui/apps/settings-app/src/components/app-routes.tsx
```diff
@@ -26,6 +26,8 @@ const AppRoutes: React.FC<RootComponentProps> = props => {
 
   const { t } = useTranslation('app-settings-ewa');
 
+  console.log(theme, '<<< the theme');
+
```
josenriagu: Do we need to log the theme or we can clean it up?

ui/hooks/src/use-login.new.ts
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Hook to check if a user is already registered
+ * @example useCheckSignup hook
+ * ```typescript
+ * const checkSignupQuery = useCheckSignup('0x42b35jk53ifq');
+ * // checkSignupQuery.data returns a boolean value which determines if one should proceed with signup process or not
+ * ```
+ */
+export function useCheckSignup(ethAddress: string) {
```
quininez: is this still being used?

quininez (approved):

PR #1971, opened 2023-06-28
Feat/topbar and toggle fixes

themonster2015:
## Checklist


ui/widgets/top-bar/src/components/Topbar.tsx
```diff
@@ -23,6 +23,17 @@ export interface ITopbarProps {
...
 }
+const notificationIcon = function (snoozeNotifications, hasNewNotifications) {
+  if (snoozeNotifications) {
+    return 'BellSnoozeIcon';
+  } else {
+    if (hasNewNotifications) {
+      return 'BellAlertIcon';
+    }
+    return 'BellIcon';
+  }
```
didd:
```suggestion
   if (snoozeNotifications) {
     return 'BellSnoozeIcon';
   } 
   return hasNewNotifications ? 'BellAlertIcon' : 'BellIcon';
```

ui/widgets/top-bar/src/components/Topbar.tsx
```diff
@@ -81,8 +92,9 @@ const Topbar: React.FC<ITopbarProps> = props => {
           onClick={onBackClick}
         />
       </div>
-      <div
-        className={tw('p-0 m-0 cursor-pointer flex(& col) justify-center items-center')}
+      <Button
+        plain={true}
+        customStyle={tw('p-0 m-0 cursor-pointer flex(& col) justify-center items-center')}
```
didd: We need to include `cursor-pointer` inside `Button` component when `plain` is true :) I should have added this.  Can we use `Stack` component for aligning and justifying center ?

ui/widgets/top-bar/src/components/Topbar.tsx
themonster2015: This is very nice :)

ui/widgets/top-bar/src/components/Topbar.tsx
themonster2015: I have updated it to use Stack. `cursor-pointer` is included as a custom-style

didd: Regarding, `cursor-pointer` is it really needed here because button's default behavior is to have a cursor pointer ? That's the part I am not clear with :)

SeverS (approved):

ui/widgets/top-bar/src/components/Topbar.tsx
themonster2015: I checked and indeed the button's default behavior is to have a cursor pointer so the added style is not really needed. But I will remove it in another PR :)

PR #1996, opened 2023-07-26
Feat: sidebar cta card

josenriagu:
## Description

- add close sidebar cta card functionality

## Checklist


ui/widgets/sidebar/src/components/cta-card.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+
+import Box from '@akashaorg/design-system-core/lib/components/Box';
+import Button from '@akashaorg/design-system-core/lib/components/Button';
+import Icon from '@akashaorg/design-system-core/lib/components/Icon';
+import Stack from '@akashaorg/design-system-core/lib/components/Stack';
+import Text from '@akashaorg/design-system-core/lib/components/Text';
+
+export type SidebarCTACardProps = {
+  ctaText: string;
```
SeverS: This is a sidebar specific component and it's not reused. I would suggest to not pass translated strings as props but directly use the `t` function inside this component. This way you will remove and cleanup the props a bit

ui/widgets/sidebar/src/components/sidebar-component.tsx
```diff
@@ -1,61 +1,81 @@
...
+    select: data => data.viewer,
+    onSuccess: data => {
+      if (data) setIsLoading(false);
+    },
+  });
 
   const sdk = getSDK();
+  const routing = plugins['@akashaorg/app-routing']?.routing;
+
+  const dismissedCardId = 'dismiss-sidebar-cta-card';
```
kenshyx: prefix with application name and remove `dismiss` from the id

ui/widgets/sidebar/src/components/sidebar-component.tsx
```diff
@@ -121,6 +131,44 @@ const SidebarComponent: React.FC<RootComponentProps> = props => {
...
+        </Text>
+      );
+    }
+  }, [isLoading, loginQuery.data?.id, t]);
+
+  useEffect(() => {
+    if (myProfileQuery.data?.profile?.did?.id) {
+      setProfileName(myProfileQuery.data?.profile?.name);
+    } else if (loginQuery.data?.id) {
+      setProfileName('Logged-in User');
```
kenshyx: why this section had to be moved up? it makes the review way harder

ui/widgets/sidebar/src/components/sidebar-component.tsx
```diff
@@ -1,61 +1,81 @@
...
+  const [profileName, setProfileName] = useState('Guest');
+
+  const subtitleUi = useRef(null);
+
   const { t } = useTranslation('ui-widget-sidebar');
 
   const loginQuery = useGetLogin();
   const logoutQuery = useLogout();
   const queryClient = useQueryClient();
+  const [dismissed, setDismissed] = useDismissedCard();
```
kenshyx: Imho the `useDismissedCard` hook should make use of a `Set` instead of `Array`. You can't dismiss multiple times the same card id, it also saves you from a lot of pain if for example a card is registered multiple times.

ui/widgets/sidebar/src/components/sidebar-component.tsx
```diff
@@ -121,6 +131,44 @@ const SidebarComponent: React.FC<RootComponentProps> = props => {
     };
   }, [routing]);
 
+  // const allApps = useMemo(() => {
+  //   return [...(worldApps || []), ...(userInstalledApps || [])];
+  // }, [worldApps, userInstalledApps]);
+
+  useEffect(() => {
```
SeverS: What is going on here? why are we storing components in ref objects? where did this come from?

ui/widgets/sidebar/src/components/sidebar-component.tsx
SeverS: this is exactly the use-case of useMemo, instead of using an internal state, compute and return a value..

josenriagu: I moved some of the existing blocks around, that's why they got highlighted. the PR primarily introduced changes relating to the CTA card

ui/hooks/src/use-dismissed-card.ts
```diff
@@ -30,15 +29,20 @@ export function useDismissedCard(
...
-  const setValue = (newClosedValue: string) => {
-    setClosed(prevClosed => {
-      if (prevClosed.includes(newClosedValue)) {
-        return prevClosed;
+  const setValue = (newId: string) => {
+    setIds(prevIds => {
+      const uniqueIds = new Set(prevIds);
+
+      if (uniqueIds.has(newId)) {
+        return [...uniqueIds];
```
kenshyx: just use the Set object, why do you have to convert it back to array?

ui/apps/settings-app/src/components/base-layout.tsx
```diff
@@ -13,7 +13,7 @@ const BaseLayout: React.FC<PropsWithChildren<IBaseLayout>> = props => {
   return (
     <BasicCardBox pad="p-0">
       <React.Fragment>
```
SeverS: This fragment is no longer needed here, you can remove it

SeverS (changes requested):
good to go after the changes

SeverS (approved):

PR #1997, opened 2023-07-27
Fix/profile fixes

themonster2015:
## Description

* Added validations for name field
* profile page and avatar for users without a profile created yet.

## Checklist


ui/design-system-components/src/components/EditProfile/GeneralForm/index.tsx
```diff
@@ -83,16 +83,20 @@ export const GeneralForm: React.FC<GeneralFormProps> = ({
...
+              <TextField
+                type="text"
+                name={name}
+                label={nameField.label}
+                value={value}
+                onChange={onChange}
+                inputRef={ref}
+              />
+              {errors.name && <p className="text-sm text-red-600 mt-1">{errors.name.message}</p>}
+            </>
```
didd:
```suggestion
              <TextField
                type="text"
                name={name}
                label={nameField.label}
                value={value}
                caption={errors.name ? errors.name.message: null}
                status={errors.name ? 'error' : null}
                onChange={onChange}
                inputRef={ref}
              />
```

We already have the props to display error on `TextField`.

ui/design-system-components/src/components/EditProfile/GeneralForm/index.tsx
```diff
@@ -83,16 +83,20 @@ export const GeneralForm: React.FC<GeneralFormProps> = ({
...
           )}
           defaultValue={nameField.initialValue || ''}
+          rules={{ required: true }}
```
didd: You can specify this using zod method https://github.com/colinhacks/zod#required on the schema, it's good to handle all validation from the schema for consistency.

ui/apps/profile/src/components/pages/profile-info-page.tsx
```diff
@@ -86,14 +108,69 @@ const ProfileInfoPage: React.FC<RootComponentProps> = props => {
...
+              }}
+              customStyle="mb-4"
+            />
+          )}
+        </Stack>
+      </>
+    );
+
+  if (validDid) {
+    return (
```
didd: This can be simplified as follows:

First, declare the following variables: 

`const hasProfile = status === 'success' && profileData;`

`const profileNotCreated = (!isLoggedIn && validDid && !hasProfile) || (isLoggedIn  && !hasProfile);` You can use this variable to conditionally display DefaultEmptyCard component.

`const profile = hasProfile ? { did : { id: profileId } } : profileData`. Then you can use this with optional chaining on `ProfileHeader` component. No need to render `ProfileHeader` using two separate conditions.

That will make the code cleaner and easier to read.

ui/apps/profile/src/components/pages/profile-info-page.tsx
```diff
@@ -16,11 +17,21 @@ import {
...
 
   const { t } = useTranslation('app-profile');
+  const sdk = getSDK();
+
+  async function resolveDid(did: string): Promise<DIDDocument> {
+    return await sdk.services.common.misc.resolveDID(did);
+  }
+
+  const [validDid, setValidDid] = React.useState(false);
+  const [isLoading, setIsLoading] = React.useState(false);
```
didd: This along with the corresponding `useEffect` needs to be moved to the hooks folder. You can create the hook as follows:

```
export function useValidDid(profileId: string, enabled?: boolean) {
     const [validDid, setValidDid] = React.useState(false);
     const [isLoading, setIsLoading] = React.useState(false);
     
     React.useEffect(() => {
       if(enabled) {
           setIsLoading(true);
           resolveDid(profileId).then(res => {
                setValidDid(Boolean(res));
                setIsLoading(false);
            });
        }
      }, [enabled]);
     
    return enabled ? {validDid, isLoading} : null;
}
```

You can use this latter as `... useValidDid(profileId, !!profileData);`

That make the code more cleaner and readable.

didd (changes requested):

didd:
Yen, I am sharing the reviews at the draft stage so that you get a chance to refactor your code before the PR is ready for review.

ui/design-system-components/src/components/EditProfile/GeneralForm/index.tsx
themonster2015: I didn't notice it :). Will check

ui/design-system-components/src/components/EditProfile/GeneralForm/index.tsx
themonster2015: I realized this condition is not needed because it already respects the rules specified in the schema :).

ui/hooks/src/use-valid-did.ts
```diff
@@ -16,11 +14,8 @@ export function useValidDid(profileId: string, enabled?: boolean) {
     if (enabled) {
       setIsLoading(true);
       resolveDid(profileId).then(res => {
-        // case when it's an eth address and we need to verify if it's a valid address separately
-        if (res && profileId.includes('eip155')) {
-          ethers.utils.isAddress(profileId.split(':').pop())
-            ? setValidDid(true)
-            : setValidDid(false);
+        if (res && 'isEthAddress' in res) {
```
kenshyx: check this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

ui/hooks/src/use-valid-did.ts
```diff
@@ -16,11 +14,8 @@ export function useValidDid(profileId: string, enabled?: boolean) {
...
+          res.isEthAddress ? setValidDid(true) : setValidDid(false);
```
kenshyx: the DID is still valid even if the `isEthAddress` is false

kenshyx: if this logic is to show the eth logo or not then please rename this

themonster2015: I think I misunderstood your point earlier. Because I thought if the isEthAddress check returns false then I would display the DID not valid page, but now I see that the DID format is still valid, so we should still display the profile page but with a broken Eth logo or something.

ui/apps/profile/src/components/pages/profile-info-page.tsx
```diff
@@ -50,9 +55,15 @@ const ProfileInfoPage: React.FC<RootComponentProps> = props => {
...
       : { isViewer: null, profile: null };
 
+  const { validDid, isLoading } = useValidDid(profileId, profileDataReq.data && !profileData);
+
   const isLoggedIn = !!loginQuery.data?.id;
 
-  if (status === 'loading') return <ProfileLoading />;
+  const hasProfile = status === 'success' && profileData;
+
+  const profile = !hasProfile ? { id: profileId } : profileData.did;
```
didd:
```suggestion
  const did = hasProfile ? profileData.did : { id: profileId };
```

ui/apps/profile/src/components/pages/profile-info-page.tsx
```diff
@@ -61,7 +72,7 @@ const ProfileInfoPage: React.FC<RootComponentProps> = props => {
     });
   }
 
-  const hasError = status === 'error' || (status === 'success' && !profileData);
+  const hasError = status === 'error'; /*  || (status === 'success' && !profileData) */
```
didd, line 75:
```suggestion
  const hasError = status === 'error'; 
```

ui/apps/profile/src/components/pages/profile-info-page.tsx
themonster2015: Thanks Didd. I have updated it.

didd:
It all looks fine to me.

didd (approved):

PR #2019, opened 2023-08-18
Feat/update profile interests page

themonster2015:
## Checklist


ui/design-system-core/src/components/Pill/index.tsx
```diff
@@ -33,7 +33,7 @@ const Pill: React.FC<PillProps> = ({
       hover={false}
       label={label}
       size={size}
-      icon={icon}
+      {...(active && { icon: icon })}
```
didd:
```suggestion
      icon={ active ? icon : null }
```

themonster2015: Thanks Didd. I was using '' and typescript complained but using null works perfectly 👍

ui/apps/profile/src/components/pages/profile-engagement/interests-page.tsx
```diff
@@ -55,7 +108,22 @@ const InterestsPage: React.FC<RootComponentProps> = props => {
     profileDataReq.data,
   );
 
-  const handleTopicClick = (topic: string) => {
+  const successText = `${t('Successfully subscribed to interest')}!`;
+
+  const handleTopicClick = topic => {
+    if (!ownTopics.find(interest => interest.value === topic.value)) {
+      createInterest
+        .mutateAsync({
```
didd: Unless you need concurrent mutations  or you absolutely need to use Promise, I don't see the reason for using mutateAsync. You will need to manually catch for errors but with mutate this isn't necessary.  So why not add the following inside the `onSuccess` method of the create interest hook ? :)
```
 setFeedbackMessage(successText);
 setShowFeedback(true);
```

ui/apps/profile/src/components/pages/profile-engagement/interests-page.tsx
```diff
@@ -75,21 +146,78 @@ const InterestsPage: React.FC<RootComponentProps> = props => {
...
+              handleClick: () => {
+                //do something
+              },
+            }}
+            saveButton={{
+              label: 'Save',
+              loading: isProcessing,
+              handleClick: interests =>
+                createInterest
+                  .mutateAsync({
```
didd: Same as the comment above for `mutateAsync`

ui/apps/profile/src/components/pages/profile-engagement/interests-page.tsx
themonster2015: You are right. It is much cleaner this way :). I was complicating things thinking I need mutateAsync to display different feedback messages to the user.

didd:
The PR looks good to me, but I will delay approving it till the pending discussion you have with Marius is resolved.

kenshyx (approved):

PR #2030, opened 2023-08-28
Chore: Cleanup old hooks dev-dashboard app

josenriagu:
## Description

- removes old hooks from dev-dashboard app
- removes redundant components and codes

## Checklist


ui/apps/dev-dashboard/src/extensions/delete-dev-key.tsx
```diff
@@ -23,27 +23,27 @@ const DeleteDevKeyModal = (props: RootExtensionProps) => {
...
-  }, [extensionData]);
-
   const keyName = React.useMemo(() => {
     if (extensionData.hasOwnProperty('keyName') && typeof extensionData.keyName === 'string') {
       return extensionData.keyName;
     }
   }, [extensionData]);
 
-  const deleteKeyMutation = useDeleteDevKey();
+  const deleteKeyMutation = {
```
SeverS, line 33: I would prefer to use a `null` value instead of a mock because it's easier to spot afterwards

SeverS: can you revisit every hook and replace that mock with null?
example:
`const myHook = null;`

josenriagu: I passed the mock so it takes into consideration other states of the hook like isSuccess, isFetching etc.

SeverS: ok but at least put a @TODO comment before each one stating that it need replacing

SeverS (approved):

PR #2048, opened 2023-09-12
Feat: update profile stats

josenriagu:
## Description

- update profile stats hook usage

## Checklist

### DIFF
--- a/ui/apps/profile/src/components/profile-stats-view.tsx
+++ b/ui/apps/profile/src/components/profile-stats-view.tsx
@@ -36,11 +36,12 @@ const ProfileStatsView: React.FC<ProfileStatsViewProps> = ({
   const sdk = getSDK();
 
   useEffect(() => {
-    const getStats = () => {
-      sdk.api.profile
-        .getProfileStats(profileId)
-        .then(({ data }) => setStats(data))
-        .catch(error => logger.error(`Error getting profile stats ${error}`));
+    const getStats = async () => {
+      const res = await sdk.api.profile.getProfileStats(profileId);
+
+      if (res.data) {
+        setStats(res.data);
+      }
     };
 
     getStats();
--- a/ui/apps/profile/src/components/profile-stats-view.tsx
+++ b/ui/apps/profile/src/components/profile-stats-view.tsx
@@ -1,8 +1,7 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 
-import getSDK from '@akashaorg/awf-sdk';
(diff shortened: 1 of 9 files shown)


ui/apps/profile/src/components/profile-stats-view.tsx
```diff
@@ -16,24 +18,33 @@ const ProfileStatsView: React.FC<ProfileStatsViewProps> = ({
...
+  const sdk = getSDK();
 
-  const stats = {
-    beamsTotal: 10,
-    interestsTotal: 10,
-    followersTotal: statsQuery.data?.akashaProfile?.followersCount,
-    followingTotal: 10,
-  };
+  useEffect(() => {
+    const getStats = () => {
```
kenshyx: I think you wanted here maybe to declare it `async` so you can use `await` instead of the promise api?

josenriagu: Okay, I will refactor it

kenshyx: you can also create a react-query hook for this call so you have caching

ui/apps/profile/src/components/profile-stats-view.tsx
```diff
@@ -21,53 +18,34 @@ const ProfileStatsView: React.FC<ProfileStatsViewProps> = ({
...
+      const res = await sdk.api.profile.getProfileStats(profileId);
 
-  const followersTotal =
-    followers.data && 'akashaProfile' in followers.data
-      ? followers.data?.akashaProfile?.followers?.edges?.length || 0
-      : 0;
+      if (res.data) {
+        setStats(res.data);
+      }
+    };
```
didd: For code resiliency we need to wrap the code block with `try ... catch`

ui/hooks/src/use-profile-stats.ts
```diff
@@ -0,0 +1,37 @@
...
+      data: {
+        totalFollowing: 0,
+        totalFollowers: 0,
+        totalBeams: 0,
+        totalTopics: 0,
+      },
+    },
+    onError: (err: Error) => logError('useProfileStats.getStats', err),
+  });
+}
```
didd, line 37: I suggest to add this line to get the query key: 

```useProfileStats.getKey = (profileId: string) => [`${PROFILE_STATS_QUERY_KEY}_${profileId}`];```. In this case we don't need to export `PROFILE_STATS_QUERY_KEY`.

didd: That approach creates abstraction around the query key so that the user of this key doesn't need to bother how it's constructed. The function `getKey` will get the desired query key. This is the approach being used within the generated hooks.

didd (approved):

PR #2074, opened 2023-10-11
chore(): migrate extends->extensions

SeverS:
## Description

Changes:
- app loader refactor
  - plugins are now loaded before anything else - no more warnings from the translation plugin or errors in layout widget regarding `*Store is not defined`
- more granular typings of `uiEvents`
- removed unused / unnecessary ui events
- navigateTo now has an optional second param `replace` that can be used to replace the current history entry (useful for modals)
- the initial redirect in a world is also replacing state - back button should work if you want to go to the previous page before loading the world
- added *stores plugin for mountable entities: (extensions, widgets and plugins) to the Extensions App
  - the Extensions App now has e greater role in a world (can be seen as part of AKASHA Core)
- added a new library of (currently React only) components to handle the stores (ui-lib-extensions)

## Checklist


ui/app-loader/src/extensions.ts
```diff
@@ -0,0 +1,37 @@
+import { WorldConfig } from '@akashaorg/typings/lib/ui';
+import getSDK from '@akashaorg/awf-sdk';
+
+export const getRemoteLatestExtensionInfos = async (extensions: { name: string }[]) => {
+  return Promise.resolve([]);
+};
+
+export const getExtensionsData = async (extNames: string[], worldConfig: WorldConfig) => {
+  const remote = extNames.filter(extName => {
+    return !worldConfig.registryOverrides.find(ext => ext.name === extName);
```
kenshyx: this can be replaced with `.some(ext => ext.name === extName)`

kenshyx (approved):

PR #2079, opened 2023-10-20
chore(): fix for sidebar connect state

themonster2015:
## Checklist


ui/widgets/sidebar/src/components/sidebar-component.tsx
```diff
@@ -203,10 +189,18 @@ const SidebarComponent: React.FC<unknown> = () => {
...
   async function handleLogout() {
-    await logoutQuery.mutateAsync();
-    await queryClient.invalidateQueries({
-      queryKey: [LOGIN_STATE_KEY],
+    setIsLoading(true);
+
+    queryClient.resetQueries({
+      queryKey: useGetMyProfileQuery.getKey(),
+    });
+    await logoutQuery.mutateAsync().then(() => {
```
kenshyx: this should make use of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race

kenshyx: the list for race contains the `mutateAsync` and the timeout functions

kenshyx: the idea is that this should not get stuck if the logoutQuery does not respond

ui/widgets/sidebar/src/components/sidebar-component.tsx
```diff
@@ -194,13 +199,16 @@ const SidebarComponent: React.FC<unknown> = () => {
...
-        setIsLoading(false);
-        // longer wait time to sign out to make sure the user is really signed out on the sdk side
-      }, 5000);
+
+    const logoutPromise = logoutQuery.mutateAsync();
+    const timeoutPromise = new Promise((resolve, _) => {
+      setTimeout(resolve, 5000);
+    });
+
+    Promise.race([logoutPromise, timeoutPromise]).then(() => {
```
kenshyx: you can await here the race to finish

ui/widgets/sidebar/src/components/sidebar-component.tsx
```diff
@@ -194,13 +199,16 @@ const SidebarComponent: React.FC<unknown> = () => {
...
+      setIsLoading(false);
```
kenshyx: you can pull everything outside of `then` if you await the race condition

ui/widgets/sidebar/src/components/sidebar-component.tsx
themonster2015: Thanks Marius.

didd:
Once you login and refresh the page there is a duration before the logged in avatar section changes on the sidebar. But in the mean time we don't see the loading placeholder. I think we need the loading placeholder otherwise it is confusing UX.

ui/widgets/sidebar/src/components/sidebar-component.tsx
```diff
@@ -55,12 +55,12 @@ const SidebarComponent: React.FC<unknown> = () => {
...
   const profileName = useMemo(
-    () => (!isLoggedIn || isLoading ? t('Guest') : myProfileQuery.data?.akashaProfile?.name || ''),
-    [myProfileQuery.data, isLoggedIn, isLoading, t],
+    () => (!isLoggedIn ? t('Guest') : myProfileQuery.data?.akashaProfile?.name || ''),
+    [myProfileQuery.data, isLoggedIn, t],
   );
 
   //empty profile name implies the connection process is still in progress
-  const inProgress = isLoading || !profileName;
+  const inProgress = isLoading || (isLoggedIn && myProfileQuery.status === 'loading');
```
didd: This line requires revision. You can refer to the comment I shared here https://github.com/AKASHAorg/akasha-core/pull/2079#issuecomment-1774711653.

themonster2015: There is a case when the user hasn't filled in their profile info page and therefore their profile name will be empty and for those users, it will keep loading forever.

didd: I understood the change you made which btw is a great catch :) But I was trying to bring to your attention an issue I observed which I shared on the comment. Here is a screen recording of it.

https://github.com/AKASHAorg/akasha-core/assets/35398733/d44d6561-0b7e-4caa-bd8d-bba751e268e7

After you login and refresh the page there is a gap before the logged in profile avatar section changes and in the mean time there is no loading placeholder. We need to revise the logic for checking `inProgress`. There is even a flicker for the loading placeholder for split seconds.

themonster2015: Hi Didd. I just made some changes. I think the issue you mentioned is fixed now...

ui/hooks/src/use-logged-in.ts
```diff
@@ -23,7 +23,7 @@ export function useLoggedIn() {
   useEffect(() => {
     if (loginQuery.data?.id && !myProfileQuery.data) {
       const refetchProfileData = async () => {
-        return myProfileQuery.refetch();
+        return (await myProfileQuery.refetch()).data;
```
kenshyx: to make this easier to maintain, something like this would work?
```ts
const { data } = await myProfileQuery.refetch();
return data
```

sdk/src/auth/index.ts
```diff
@@ -347,7 +347,7 @@ class AWF_Auth {
       return Promise.resolve(null);
     }
     const localUser = localStorage.getItem(this.currentUserKey);
-    if (localUser) {
+    if (localUser && !this._lockSignIn) {
```
kenshyx: The lock happens after this block, it shouldn't have any effect this change.

kenshyx: Actually this would work only for multiple `getCurrentUser()` requests in paralel

didd: On some refresh,  automatic sign in doesn't work and the login placeholder keeps on loading. It seems that the automatic sign in triggered via calling `getCurrentUser` never emits SIG_IN event.  It does appear that the  `eth_requestAccounts` request to MetaMask is never fulfilled and the lines after this request never executes. 

It has been quite tough to reproduce the issue being observed on the PR review locally and the best thing now is add some console logs and debug from the PR previews.

ui/hooks/src/use-logged-in.ts
```diff
@@ -10,8 +12,50 @@ import { useGetLogin } from './use-login.new';
...
+      console.log('first condition true');
+
+      const refetchProfileData = async () => {
+        const { data } = await myProfileQuery.refetch();
+        return data;
+      };
+
+      if (!refetchProfileData()) {
+        queryClient.setQueryData([LOGIN_STATE_KEY], null);
+      }
```
didd:
```suggestion
      const refetchProfileData = async () => {
            await myProfileQuery.refetch();
      };
      
      refetchProfileData();
```

I am not seeing the need to do `queryClient.setQueryData([LOGIN_STATE_KEY], null);`. And if this is needed then refetchProfileData() is still a promise as you have wrote it. Either you need an `IIFE` or move the `setQueryData` logic inside the `refetchProfileData` function.

ui/hooks/src/use-logged-in.ts
```diff
@@ -10,8 +12,50 @@ import { useGetLogin } from './use-login.new';
...
+      console.log('second condition true');
+
+      const resetQueryPromise = queryClient.resetQueries(useGetMyProfileQuery.getKey());
+      const timeoutPromise = new Promise((resolve, _) => {
+        setTimeout(resolve, 3000);
+      });
+
+      const res = async () => Promise.race([resetQueryPromise, timeoutPromise]);
+      res();
+    }
```
didd: :D TBH I am not sure what exactly is going on here. I thought the use of the promise race is supposed to be between resetting useGetMyProfileQuery query and the logout mutation as the logout may not respond ? 

And btw on line 47 I don't see the use of async as you aren't awaiting anything.

ui/hooks/src/use-logged-in.ts
```diff
@@ -10,8 +12,50 @@ import { useGetLogin } from './use-login.new';
...
+  }, [loginQuery.data, myProfileQuery, queryClient]);
+
+  return {
+    isLoggedIn: !!myProfileQuery.data,
+    loggedInProfileId: (
+      myProfileQuery.data?.akashaProfile?.did.id ??
+      (loginQuery.data?.id || '')
+    ).toLowerCase(),
```
didd, line 44: It is beneficial to add a comment here on why you chose to obtain data from myProfileQuery and not loginQuery. It seems like you want to first wait the result from my profile query to finish so that you can display the avatar and the profile id at the same time. Otherwise your intent may not be straightforward just by looking at the code :)

ui/hooks/src/use-logged-in.ts
themonster2015: What I was thinking was to wait some time for the resetQueries to finish. Since it doesn't return anything when it finishes, I have no way of knowing when it finishes or if it ran successfully. So I use the timeout function to leave some time for it.

ui/hooks/src/use-logged-in.ts
themonster2015: I use this one `queryClient.setQueryData([LOGIN_STATE_KEY], null)` for the case when I click on connect immediately after logging out and the `LOGIN_STATE_KEY` just switched from null to logged in data automatically.  But I'm removing this entire logic and testing if it still works or not on the PR preview.

didd: The refetch is needed but not `queryClient.setQueryData([LOGIN_STATE_KEY], null)`.

didd (dismissed):

ui/widgets/sidebar/src/components/sidebar-component.tsx
```diff
@@ -182,11 +182,11 @@ const SidebarComponent: React.FC<unknown> = () => {
     }
   }
 
-  async function handleLogout() {
+  function handleLogout() {
     setIsLoading(true);
 
-    await logoutQuery.mutateAsync();
     queryClient.setQueryData([LOGIN_STATE_KEY], null);
+    logoutQuery.mutate();
```
didd: nitpick: logoutQuery -> logoutMutation

didd (approved):

PR #2081, opened 2023-10-26
Feat/lazy loaders

themonster2015:
## Description

* Add lazy loaders

## Checklist


ui/lib/feed/src/components/beam-feed.tsx
```diff
@@ -77,7 +78,7 @@ const BeamFeed: React.FC<BeamFeedProps> = props => {
     <>
       {!beamsReq.initialScrollState.isFetched && (
         <Stack fullWidth={true} customStyle="p-8">
```
SeverS: the p-8 here is the reason for layout shifting on widgets, you can safely remove it because you added a full width element (AntennaLoader)

themonster2015: Thanks, Sever!

SeverS: And the main reason why it happens is because you have some fixed widths on the skeletons (on the lines) so the browser tries to keep the specified width when applying the padding. It is better to use percentages when defining the widths of those lines

ui/design-system-components/src/components/Loaders/antenna-loader.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </Stack>
+      </Card>
+      {Array.from({ length: 4 }).map(() => (
+        <Card radius="rounded-2xl">
+          <Stack direction="column" align="start" justify="center" spacing="gap-y-4">
+            <Stack direction="row" align="start" justify="between" fullWidth>
+              <Stack direction="row" align="center" justify="center" spacing="gap-x-1">
+                <CircularPlaceholder height="h-10" width="w-10" customStyle="shrink-0" animated />
+                <Stack direction="column" align="center" justify="between" spacing="gap-y-1">
+                  <TextLine title="tagName" animated={true} width="w-[147px]" />
```
SeverS: here instead `w-[147]px` you should use percent. In this case it's not very important to have exactly the same width as in the design files but it will help in case there is some padding added to this skeleton

ui/design-system-components/src/components/Loaders/sidebar-loader.tsx
```diff
@@ -0,0 +1,123 @@
...
+        <Stack direction="row" customStyle="w-full h-fit justify-between">
+          <Stack direction="row" spacing="gap-x-1">
+            <CircularPlaceholder height="h-10" width="w-10" customStyle="shrink-0" animated />
+            <Stack
+              direction="column"
+              spacing="gap-y-1"
+              customStyle="max-w(xl:[10rem] lg:[8rem] md:[10rem] xs:[2rem])"
+            >
+              <TextLine title="tagName" animated={true} width="w-[147px]" />
+              <TextLine title="tagName" animated={true} width="w-[147px]" />
```
SeverS: Same here, percent instead px

ui/widgets/trending/src/components/trending-widget-component.tsx
```diff
@@ -96,6 +97,32 @@ const TrendingWidgetComponent: React.FC<unknown> = () => {
       getNavigationUrl: navRoutes => `${navRoutes.Tags}/${topic}`,
     });
   };
+  React.useEffect(() => {
+    if (needToProcessTags.length > 0) {
+      while (needToProcessTags.length > 0) {
+        console.log('needToProcessTags', needToProcessTags);
```
SeverS: rm console.log

didd (approved):

PR #2082, opened 2023-10-31
chore(): edit icon only on own beams

themonster2015:
## Description

* Display edit icon only on own beam cards.

## Checklist

### DIFF
--- a/ui/apps/lists/src/components/lists-page.tsx
+++ b/ui/apps/lists/src/components/lists-page.tsx
@@ -11,7 +11,6 @@ import ListAppTopbar from '@akashaorg/design-system-components/lib/components/Li
 import DefaultEmptyCard from '@akashaorg/design-system-components/lib/components/DefaultEmptyCard';
 import { EntityTypes, ModalNavigationOptions } from '@akashaorg/typings/lib/ui';
 import { useLoggedIn, useRootComponentProps } from '@akashaorg/ui-awf-hooks';
-import { useGetMyProfileQuery } from '@akashaorg/ui-awf-hooks/lib/generated/hooks-new';
 
 const ListsPage: React.FC<unknown> = () => {
   const [showModal, setShowModal] = React.useState(false);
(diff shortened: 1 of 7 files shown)


ui/design-system-components/src/components/Entry/EntryCard/index.tsx
```diff
@@ -146,13 +146,11 @@ const EntryCard: React.FC<EntryCardProps> = props => {
...
           {entryData?.createdAt &&
-            (!editable && notEditableLabel ? (
+            (editable ? (
               <Tooltip placement={'top'} content={notEditableLabel}>
                 {editableIconButtonUi}
               </Tooltip>
-            ) : (
-              editableIconButtonUi
-            ))}
+            ) : null)}
```
didd: That logic was there for a reason :) Can you share the reason why removed it ? I agree `!editable` check isn't needed but the rest are important.

The logic was there for two reasons. One, when `notEditableLabel` is set then it adds unnecessary tooltip element into a page and two when you hover on the `edit` icon then you will see unwanted arrow.

image

themonster2015: thanks for pointing it out Didd :). I overlooked the notEditableLabel prop. Will take a closer look.

ui/apps/lists/src/components/lists-page.tsx
```diff
@@ -21,15 +21,16 @@ const ListsPage: React.FC<unknown> = () => {
 
   const bookmarkDelete = null;
 
+  const { isLoggedIn } = useLoggedIn();
+
   const profileDataReq = useGetMyProfileQuery(null, {
```
didd: We can use `loggedProfileData` from `useLoggedIn` hook instead of recomputing it here :)

themonster2015: This component needs updating yes

ui/lib/feed/src/components/cards/beam-card.tsx
```diff
@@ -22,10 +22,11 @@ type BeamCardProps = Pick<
   | 'disableActions'
 > & {
   entryData: AkashaBeam;
+  loggedProfileData: Profile;
```
didd, line 25: This prop isn't needed. We can use the `isViewer` flag from the author object to decide whether to show or hide the edit button. As a result you don't need to set the editable prop for such purpose. 

Btw there is a new design I have been working on that of placing the edit button within an ellipsis menu and I have taken care of all this on the PR #2086.

PR #2087, opened 2023-11-02
Fix race condition issue in Latest topic widget

themonster2015:
## Description

* Fix race condition issue in Latest topic widget 

## Checklist

### DIFF
--- a/ui/widgets/trending/src/components/trending-widget-component.tsx
+++ b/ui/widgets/trending/src/components/trending-widget-component.tsx
@@ -149,6 +149,7 @@ const TrendingWidgetComponent: React.FC<unknown> = () => {
           })();
           break;
         case 'unsub':
+          if (!tagSubscriptions.find(tag => tag.value === currentTag?.topic)) return;
           (async () => {
             await createInterestMutation.mutateAsync({
               i: {
--- a/ui/widgets/trending/src/components/trending-widget-component.tsx
+++ b/ui/widgets/trending/src/components/trending-widget-component.tsx
@@ -15,7 +15,7 @@ import {
   useCreateInterestsMutation,
   useGetFollowDocumentsQuery,
 } from '@akashaorg/ui-awf-hooks/lib/generated/hooks-new';
-import { useQueryClient } from '@tanstack/react-query';
+import { useQueryClient, useIsMutating } from '@tanstack/react-query';
 
 import Stack from '@akashaorg/design-system-core/lib/components/Stack';
 import ErrorLoader from '@akashaorg/design-system-core/lib/components/ErrorLoader';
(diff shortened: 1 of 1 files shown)


didd:
Another approach I suggest  is to create a self contained component for each tag row. This component will contain the tags, the subscribe button and the hooks to subscribe and unsubscribe. In that manner you don't need to worry about this race condition as each tag component manages it's own subscription/un-subscription. What do you think of this approach?

themonster2015:
> Another approach I suggest is to create a self contained component for each tag row. This component will contain the tags, the subscribe button and the hooks to subscribe and unsubscribe. In that manner you don't need to worry about this race condition as each tag component manages it's own subscription/un-subscription. What do you think of this approach?



> Another approach I suggest is to create a self contained component for each tag row. This component will contain the tags, the subscribe button and the hooks to subscribe and unsubscribe. In that manner you don't need to worry about this race condition as each tag component manages it's own subscription/un-subscription. What do you think of this approach?

Can you explain a bit more Didd? Because from what I understand, the way the useCreateInterestsMutation works is that it modifies the entire list of interests of a user whenever it runs, so if I implement self-contained components for each tag, and then click on subscribe or unsubscribe fast, one mutation may still try to modify a stale state because the other mutation has just finished modifying it.

didd:
> > Another approach I suggest is to create a self contained component for each tag row. This component will contain the tags, the subscribe button and the hooks to subscribe and unsubscribe. In that manner you don't need to worry about this race condition as each tag component manages it's own subscription/un-subscription. What do you think of this approach?
> 
> > Another approach I suggest is to create a self contained component for each tag row. This component will contain the tags, the subscribe button and the hooks to subscribe and unsubscribe. In that manner you don't need to worry about this race condition as each tag component manages it's own subscription/un-subscription. What do you think of this approach?
> 
> Can you explain a bit more Didd? Because from what I understand, the way the useCreateInterestsMutation works is that it modifies the entire list of interests of a user whenever it runs, so if I implement self-contained components for each tag, and then click on subscribe or unsubscribe fast, one mutation may still try to modify a stale state because the other mutation has just finished modifying it.

Sure, so inside `latest-topics` you can have a state to manage the subscribed tags. The initial value of this state can be set as `subscribedTags`. Then each of the tag component can receive this as a prop and you subscribe/unsubscribe based on this prop. So whenever you subscribe or unsubscribe you modify this state as well with a modifier function prop. So using this approach each mutation will be based on the latest state of subscribed topics. What do you think ?


One more point, you can manage `isLoading` state for each tag component to keep track of the in progress mutation state. You can set it to true `onMutate` and false on `onSettled` handlers of create mutation hook.

kenshyx (approved):

PR #2091, opened 2023-11-09
Fix/minor fixes

themonster2015:
## Checklist


ui/apps/search/src/components/routes/onboarding-page.tsx
```diff
@@ -52,23 +48,23 @@ const OnboardingPage: React.FC<OnboardingPageProps> = props => {
   };
 
   const toggleTagSubscribe = (tagName: string) => {
-    if (!isLoggedIn) {
+    if (!loggedProfileData) {
```
didd: You are only checking if user has logged in, you aren't using the profile data anywhere on the file :) `isLoggedIn` from the `useLoggedIn` hook makes sense here. Perhaps you can fix this on your other PR where you refactored the hook.

ui/design-system-core/src/components/DuplexButton/index.tsx
```diff
@@ -44,6 +44,7 @@ const DuplexButton = (props: DuplexButtonProps) => {
   const [iconOnly, setIconOnly] = useState(window.matchMedia('(max-width: 992px)').matches);
   const activeHoverIconElem = activeHoverIcon ?? icon;
   const activeIconElem = activeIcon ?? icon;
+  const fixWidth = 'w-[7rem]';
```
didd: How about setting a minimum width here rather than a fixed width or how about using a fixedWidth prop to make a conscious decision to use fixed width ? Otherwise if at some point we have a `DuplexButton` that needs to extend beyond `7rem` then this wont work unless we want to wrap text.

didd (approved):

PR #2109, opened 2023-11-29
add loading state when checking if current session exists

themonster2015:
## Description

This PR adds a loading state in the `useGetLogin` hook while it is checking if a current session exists with `getCurrentUser` so that the EditProfile component can use this `isLoading` state to know when the hook is still checking and when it actually finishes and then it can redirect accordingly if the user is not logged in.

## Checklist

### DIFF
--- a/ui/hooks/src/use-login.new.ts
+++ b/ui/hooks/src/use-login.new.ts
@@ -72,7 +72,7 @@ export function useGetLogin(onError?: (error: Error) => void) {
     },
   });
 
-  return { data: loginData, isLoading: isLoading };
+  return { data: loginData, isLoading };
 }
 
 /**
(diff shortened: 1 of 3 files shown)


josenriagu (dismissed):

ui/hooks/src/use-login.new.ts
```diff
@@ -40,14 +40,16 @@ export function useConnectWallet(provider: EthProviders) {
...
 export function useGetLogin(onError?: (error: Error) => void) {
-  const [loginData, setLoginData] = useState(null);
+  const [loginData, setLoginData] = useState<CurrentUser>(null);
+  const [isLoading, setIsLoading] = useState(true);
 
   // check previous session
   useEffect(() => {
     const getCurrentUser = async () => {
       const sdk = getSDK();
       const data = await sdk.api.auth.getCurrentUser();
```
SeverS: It might be great to add some error handling here, try/catch at least to avoid error propagation

ui/hooks/src/use-login.new.ts
```diff
@@ -67,7 +69,7 @@ export function useGetLogin(onError?: (error: Error) => void) {
     },
   });
 
-  return { data: loginData };
+  return { data: loginData, isLoading: isLoading };
```
didd:
```suggestion
  return { data: loginData, isLoading };
```

didd (approved):

PR #2112, opened 2023-12-03
feat(): switch hooks on profile app to apollo v1 

didd:
## Description

- Switch hooks on edit page, info page and follow profile button to apollo
- Clean up entry and media utils files
- Set up getMediaUrl prop on beam and reflection cards
- Fix author profile info on pending reflection card

## Checklist


ui/apps/profile/src/components/profile-header/follow-button.tsx
```diff
@@ -0,0 +1,41 @@
...
+type FollowButtonProps = {
+  profileID: string;
+  showLoginModal: (redirectTo?: { modal: ModalNavigationOptions }) => void;
+};
+
+const FollowButton: React.FC<FollowButtonProps> = props => {
+  const { profileID, showLoginModal } = props;
+  const { isLoggedIn } = useLoggedIn();
+  const sdk = getSDK();
+  const { data, error } = useGetFollowDocumentsSuspenseQuery({
```
kenshyx: why you need here the composeDB context? This should be used only for mutations

didd: I was just about to raise this point on the implementation details channel. I am receiving empty data and out of curiosity added this and I begun seeing data.

ui/apps/profile/src/components/profile-header/withProfileHeader.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { ReactNode, Suspense } from 'react';
+import FollowingFeedback from '../following-feedback';
+import ProfileHeaderView, { ProfileHeaderViewProps } from './profile-header';
+import { ProfileHeaderLoading } from '@akashaorg/design-system-components/lib/components/Profile';
+
+export const withProfileHeader = <T extends ProfileHeaderViewProps>(
```
SeverS: is this HOC used somewhere? I'm curious why you chose to create a HOC instead of normal composition using children

ui/hooks/src/utils/media-utils.ts
```diff
@@ -172,19 +124,19 @@ export const saveMediaFile = async ({ name, content, isUrl, email }: ISaveMediaF
 };
 
 /**
- * Utility to get profile image versions with media url
+ * Utility to get profile image url
  */
-export const getProfileImageVersionsWithMediaUrl = (
+export const getProfileImageUrl = (
```
SeverS: can't this function be applied to all images? is profile image different?

ui/apps/profile/src/components/profile-header/withProfileHeader.tsx
didd: In this case it's sole purpose is to make the code readable. It's used inside `app-routes` like this: 
```
  <Route
              path={`:profileId${menuRoute[FOLLOWING]}`}
              element={withProfileHeader(<FollowingPage showLoginModal={showLoginModal} />)(
                commonHeaderViewProps,
              )}
            />
  ```

File: ui/hooks/src/utils/media-utils.ts (hunk 3 above)
Comment (didd) [reply ui/hooks/src/utils/media-utils.ts]: The parameter it takes and the one it returns are specifically related to the profile image. We are working with this type `AkashaProfileImageVersions`.

File: ui/apps/profile/src/components/profile-header/withProfileHeader.tsx (hunk 2 above)
Comment (SeverS) [reply ui/apps/profile/src/components/profile-header/withProfileHeader.tsx]: hmm, I don't think it makes it more readable, imagine a component that has many more props:

```
withProfileHeader(
    <SomeComponent
       with={many}
       more={props}
    />
)(
  someOtherProps
)
```
do you think is more readable than
```
  <ProfileHeader someProps={}>
    <SomeComponent
       with={many}
       more={props}
    />
  </ProfileHeader>
  ```
  ?

didd:
```
<ProfileHeader someProps={}>
    <SomeComponent
       with={many}
       more={props}
    />
  </ProfileHeader>
  ```
  
  Profile header containing siblings as children is something I wasn't able to wrap my head around that's why I created the HOC :D And that's what I meant by making it more readable.

Comment (SeverS) [reply ui/apps/profile/src/components/profile-header/withProfileHeader.tsx]: using fragments like:

```
<>
  <ProfileHeader />
  <OtherComponent />
</>
```
or actually modifying ProfileHeader to accept children (because it suppose to be a header)

SeverS: The point i'm raising here is that you are mixing the patterns:
I would expect a HOC to be called like `withHeader(SomeComponent)` but you are using as it's a parent component `withHeader(<SomeComponent />)(someProps)`

ui/hooks/src/utils/media-utils.ts
SeverS: The question is do we have to create a function for every typing? Can we use function overloading? is the return type different for other images?

ui/apps/profile/src/components/profile-header/withProfileHeader.tsx
didd: I have created a component named as `ProfileWithHeader` to settle this :)

ui/hooks/src/utils/media-utils.ts
didd: We have a generic `getMediaUrl` and `getProfileImageUrl` is specific to profile images. I don't see the need for function overloading here. The function serves as a utility to transform a profile image version into one that profile related components can consume. 

I have renamed `getProfileImageUrl` to `transformImageVersions`.

didd: > is the return type different for other images?

Yes

SeverS (approved):

PR #2152, opened 2024-01-11
 tag feed page + nsfw content dispay/hide based on user's authenticated state

themonster2015:
## Description

* Create a tag feed page that will be displayed when clicking on each tag in the widget
* Switch to BeamStream hook for use-beams hook
* Add logic to hide nsfw content for not logged-in users.

## Checklist


ui/hooks/src/use-tags.ts
```diff
@@ -0,0 +1,212 @@
+import React from 'react';
```
quininez: I would rename this file to use-beams-by-tag, to better reflect what it actually does

themonster2015: I agree. It's a more suitable name

quininez (approved):

PR #2153, opened 2024-01-12
Refactor: beam editor tags

josenriagu:
## Description

- improve beam editor add tags flow and design

## Checklist


themonster2015:
After adding some tags and going back to the main beam editor, if I wanted to go back and delete some tags, I wasn't able to click on the Save button because it remained disabled, but after clicking on Cancel and going back to the main editor, the tags actually got removed. It was confusing.

themonster2015:
The delete flow works fine now 👍

ui/apps/akasha/src/extensions/beam-editor/beam-editor.tsx
```diff
@@ -66,6 +70,18 @@ export const BeamEditor: React.FC = () => {
     }
   }, [blocksInUse]);
 
+  React.useEffect(() => {
```
quininez: why not do the checks from this useEffect in the addTag function?

josenriagu: Alright. Let me check also in the keyup handler

quininez (approved):

PR #2157, opened 2024-01-18
Refactor/virtual list

SeverS:
## Checklist


ui/hooks/src/use-pending-reflections.ts
```diff
@@ -0,0 +1,26 @@
+import { makeVar, useReactiveVar } from '@apollo/client';
+import type { ReflectEntryData } from '@akashaorg/typings/lib/ui';
+
+const pendingReflectionsVar = makeVar([]);
```
kenshyx: This being global doesn't introduce weird bugs when adding reflections to different beams? I'm thinking about using an init function for this, wouldn't this work?

kenshyx: or initialize this at the component level and pass it to the `usePendingReflections`

SeverS: ok so it should be better to reset it on navigation rather than dangling around.
the only 'downside' will be in case if the publishing takes a bit longer and you navigate from and immediately back to it. but i guess this is quite an edge case

SeverS: made the change and it doesn't work anymore...

kenshyx (approved):

PR #2163, opened 2024-01-26
Nsfw settings app

themonster2015:
## Description

* Add NSFW option in the Settings app and apply the setting in the antenna app and profile beam page.
* Apply NSFW settings on the block level
* Modify the editor so that when there is at least one block marked as NSFW, the whole beam is NSFW.

## Checklist


ui/apps/akasha/src/components/pages/entry-page/beam-page.tsx
```diff
@@ -43,6 +72,21 @@ const BeamPage: React.FC<unknown> = () => {
     navigateToModal({ name: 'login' });
   };
 
+  const showNsfwCard = React.useMemo(() => {
+    if (
+      moderationData === AkashaBeamStreamModerationStatus.Nsfw &&
```
kenshyx, line 98: you can return directly this check

ui/apps/akasha/src/components/pages/profile-feed-page/profile-feed-page.tsx
```diff
@@ -67,6 +67,8 @@ const ProfileFeedPage = (props: ProfilePageProps) => {
                       `${navRoutes.Beam}/${itemData.node.id}${navRoutes.Reflect}`,
                   })
                 }
+                showNSFWCard={false}
+                isLoggedIn={true}
```
kenshyx: hmm, why is this default true? O_O

ui/hooks/src/use-beams.ts
```diff
@@ -237,21 +246,43 @@ export const useBeams = ({ overscan, filters, sorting, did }: UseBeamsOptions) =
...
       const initialVars: GetBeamStreamQueryVariables & GetBeamsByAuthorDidQueryVariables = {
+        ...mergedVars,
         sorting: { createdAt: SortOrder.Desc },
         id: did ?? undefined,
         indexer: did ? undefined : sdk.services.gql.indexingDID,
       };
 
+      if (!showNsfw || !isLoggedIn) {
+        initialVars.filters = {
+          and: [{ where: { status: { isNull: true } } }],
```
kenshyx: you don't need `and` here if there is just 1 condition

ui/hooks/src/use-beams.ts
```diff
@@ -237,21 +246,43 @@ export const useBeams = ({ overscan, filters, sorting, did }: UseBeamsOptions) =
...
+        };
+      }
+
+      if (showNsfw && isLoggedIn) {
+        initialVars.filters = null;
```
kenshyx: this does not follow what we had discussed last time, it should have been an `or` filter; what has changed since our discussion?

themonster2015: For this one, my intention was to use no filter when the user is logged in and allows NSFW content.

kenshyx: Yes but the problem is that there are multiple status levels(ex: suspended, removed, etc) which would be loaded if I check nsfw.

ui/apps/akasha/src/components/pages/profile-feed-page/profile-feed-page.tsx
themonster2015: This one I remember putting it to default to true because the profile feed page can only be accessed by logged-in users....

kenshyx: Isn't it safer to use the hook for checking if the user is logged in here and pass that value? Otherwise we will end up with weird racing conditions.

ui/hooks/src/use-beams.ts
kenshyx: also the condition is not ok, for example the status `OK` should be loaded

themonster2015: because right now in the data returned, there are only two statuses: null and NSFW, I think I will use the `or` condition here to include both null and OK to get the correct result :)

kenshyx: Yes but we should follow the schema not the data.

kenshyx (approved):

PR #2164, opened 2024-01-29
Feat: Vibes app

josenriagu:
## Description

- update vibe app

## Checklist


ui/apps/profile/package.json
```diff
@@ -45,14 +45,12 @@
...
     "react-i18next": "12.2.0",
     "@tanstack/react-query": "4.29.3",
-    "react-router-dom": "6.3.0",
     "single-spa-react": "6.0.1",
     "@tanstack/react-router": "1.16.6"
   },
   "devDependencies": {
     "@akashaorg/af-testing": "*",
-    "@akashaorg/typings": "*",
-    "@tanstack/router-devtools": "1.16.6"
```
quininez, line 55: why remove tanstack devtools here?

ui/apps/vibes/src/utils/format-date.ts
```diff
@@ -0,0 +1,11 @@
+import dayjs from 'dayjs';
```
quininez: you can add this to the time.ts in design-system-core, so we have all the date formatting in one place

quininez (changes requested):

ui/apps/profile/package.json
josenriagu: I have moved the devtools to topbar so it stays on for very route, so it is now redundant on the profile app

typings/src/ui/router.ts
```diff
@@ -0,0 +1,12 @@
+import { useApolloClient } from '@apollo/client';
+
+export type ApolloClient = ReturnType<typeof useApolloClient>;
```
didd: No need to export this. We can use the class `ApolloClient` from the library as type `ApolloClient<object>`.
`ReturnType<typeof useApolloClient>` is the same as `ApolloClient<object>`

quininez (approved):

PR #2167, opened 2024-02-02
feat(): implement content block nsfw handler

didd:
## Description

- add eighteen plus icon placeholder
- implement content block nsfw handler

## Checklist


ui/apps/akasha/src/extensions/beam-editor/beam-editor.tsx
```diff
@@ -72,6 +75,17 @@ export const BeamEditor: React.FC = () => {
 
   const handleNsfwCheckbox = () => {
     setIsNsfw(!isNsfw);
+
+    if (!isNsfw) {
+      range(0, blocksInUse.length).forEach(key => {
+        setNsfwBlocks(new Map(nsfwBlocks.set(key, true)));
```
kenshyx: Can you please explain the logic in this loop?

didd: It sets all content blocks on the editor as nsfw blocks when the nsfw checkbox on the header(beam) is set. 
image

kenshyx: Why use `range` function from lodash instead of a normal for loop?

kenshyx: also is there a need to create n*Map objects where `n` is the number of blocks? can you please explain the logic?

didd: It's meant to make it declarative and read better but I see from your comment now that may not be  what we have here. I will revert to the normal for loop.

kenshyx: There are also fundamental differences between a forEach function and a for loop. You can never exit from a forEach.

didd: Indeed

kenshyx: also the overhead of that range function from lodash is crazy https://github.com/lodash/lodash/blob/main/src/.internal/baseRange.ts

didd: It's not worth it. Initially I considered the overhead as well but gave weight to readability.

didd: > also is there a need to create n*Map objects where n is the number of blocks? can you please explain the logic?

It's to make react aware that I have a state change and for it to cause a re-render. That's why I am doing this :

`setNsfwBlocks(new Map(nsfwBlocks.set(key, true)))`

kenshyx: ok but couldn't this be done in one update using just 1 Map object?

didd: You are right. We can do that. :D That one map object is enough indeed.

didd: Thanks!

kenshyx (approved):

PR #2172, opened 2024-02-13
feat(): implement reflections count

didd:
## Description

-  implement reflections count

## Checklist

### DIFF
--- a/ui/hooks/src/utils/entry-utils.ts
+++ b/ui/hooks/src/utils/entry-utils.ts
@@ -110,7 +110,10 @@ export const mapReflectEntryData = (
  * Utility to map beam entry data
  */
 export const mapBeamEntryData = (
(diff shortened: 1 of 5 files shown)


ui/hooks/src/utils/entry-utils.ts
```diff
@@ -126,5 +129,6 @@ export const mapBeamEntryData = (
     tags: beam?.tags
       ?.filter(labeledTag => labeledTag.labelType === sdk.services.gql.labelTypes.TAG)
       .map(labeledTag => labeledTag.value),
+    reflectionsCount: beam?.reflectionsCount,
```
kenshyx, line 132: this number includes all reflections no matter of their state(ex: removed) so it's not accurate all the time

kenshyx: we need to track these cases but it should be enough atm

kenshyx (approved):

PR #2180, opened 2024-02-23
feat(): enhance content block extension

didd:
## Description

- divide content block extension component into read and edit components 
- adjust the style of error and app not installed fallback cards
- show display name of content block instead of the property type

## Checklist


ui/lib/extensions/src/react/content-block/read-content-block.tsx
```diff
@@ -0,0 +1,217 @@
...
+      !state.isMatched
+    ) {
+      resolveConfigs({ matchingBlocks, mode: ContentBlockModes.READONLY })
+        .then(newBlocks => {
+          setState({
+            parcels: newBlocks,
+            isMatched: true,
+          });
+        })
+        .catch(err => console.error('failed to load content blocks', err));
```
kenshyx: I saw this approach in multiple places, there is no error handling, this is just capturing the error. Also for logging the error we should use a proper logger, not the console, I remember we also used to pass as a prop the logger instance.

didd: I have addressed the issue on this PR and elsewhere where it's straightforward to replace console with logger.

didd: I can create an issue on space for error handling from UX perspective.

didd: I have addressed the console.error issue for files on this PR and elsewhere where it's straightforward to replace console with logger.

I think we can improve error handling from UX perspective. For example, show users meaningful error message card. Introduce retry and the like.

I suggest we discuss this on the daily.

ui/apps/akasha/src/extensions/beam-editor/beam-editor.tsx
```diff
@@ -277,13 +275,10 @@ export const BeamEditor: React.FC = () => {
...
                   isNsfwCheckboxSelected={!!nsfwBlocks.get(idx)}
                 />
-                <ContentBlockExtension
-                  editMode={{
-                    appName: block.appName,
-                    propertyType: block.propertyType,
-                    externalHandler: value => updateBlockDisablePublishState(value, block.order),
-                  }}
-                  mode={ContentBlockModes.EDIT}
+                <EditContentBlockExtension
```
SeverS: I think we should rename this components to EditorBlockExtension and ContentBlockExtension for better understanding their purpose

SeverS (approved):

PR #2183, opened 2024-02-26
switch to helmet-async

themonster2015:
## Description

* Switch to react-helmet-async

## Checklist


quininez (changes requested):
I think we discussed that instead of wrapping each app individually with StrictMode, you can add it just to the provider-hoc.tsx component, and this one already wraps all the apps, so it will propagate from there

themonster2015:
> I think we discussed that instead of wrapping each app individually with StrictMode, you can add it just to the provider-hoc.tsx component, and this one already wraps all the apps, so it will propagate from there

I tried that approach but encountered an issue similar to what Marius mentioned where the sidebar, topbar, and widgets are all in loading state forever. I did some research and found that single-spa Parcel has an existing problem with StrictMode https://github.com/single-spa/single-spa-react/issues/160 and we use Parcel in the Widget extension. So if we wrap all the apps here, it will include all the widgets and cause that issue.

quininez (approved):

PR #2196, opened 2024-03-06
Display toolbars only on focused block

quininez:
## Description
* provide block instances with a method to set their focused state

## Checklist

### DIFF
--- a/ui/apps/akasha/src/extensions/beam-editor/beam-editor.tsx
+++ b/ui/apps/akasha/src/extensions/beam-editor/beam-editor.tsx
@@ -41,11 +41,16 @@ export const BeamEditor: React.FC = () => {
   });
 
   const [uiState, setUiState] = useState<uiState>('editor');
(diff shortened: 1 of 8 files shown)


ui/apps/akasha/src/extensions/beam-editor/beam-editor.tsx
```diff
@@ -259,7 +273,14 @@ export const BeamEditor: React.FC = () => {
       <Stack customStyle="relative h-full overflow-hidden">
         <Stack customStyle="overflow-auto h-full">
           {blocksInUse.map((block, idx) => (
-            <div key={`${block.key}`} id={`${block.propertyType}-${idx}`}>
+            <div
```
didd: Btw the `Card` component has a prop named `type` and a value `plain` which does the same thing as what you have on the div.

quininez: ok, I'll try it out

didd (approved):

PR #2199, opened 2024-03-07
Feat: Vibes Console [Applications Center]

josenriagu:
## Description

- vibes console applications center flows

## Checklist


ui/apps/vibes-console/src/components/app-routes/index.tsx
```diff
@@ -1,29 +1,88 @@
...
+  MY_APPLICATIONS,
+  MY_APPLICATION_DETAIL,
+  WITHDRAW_APPLICATION,
+} from '../../routes';
 
 const rootRoute = rootRouteWithContext()({
   component: Outlet,
 });
 
 const applicationsRoute = new Route({
```
quininez: Route is deprecated, should be replaced with createRoute, also rootRouteWithContext is deprecated [see reference](https://tanstack.com/router/latest/docs/framework/react/api/router)

ui/apps/vibes-console/src/utils/dummy-data.ts
```diff
@@ -0,0 +1,114 @@
+import { randomDateBetweenValues } from '@akashaorg/design-system-core/lib/utils';
```
quininez: remove dummy data from apps, keep it for stories

ui/design-system-core/src/utils/time.ts
```diff
@@ -47,4 +47,11 @@ const formatRelativeTime = (date: string, locale?: string) => {
   return '';
 };
 
-export { formatDate, formatDateShort, formatRelativeTime };
+const randomDateBetweenValues = (start = 'Jan 01 2020', end = 'Dec 31 2020') => {
```
quininez: if this is used for dummy data, remove it from here

quininez (changes requested):

quininez:
can you also add react-helmet to this app? you can add it in the app.tsx, check for example extensions app

ui/design-system-core/src/utils/time.ts
josenriagu: You suggested earlier that I move that method to time utils in DS core. could be used any other place in future

quininez: it seems very specific for the vibes app, even if I suggested this, which I don't recall, I consider that it's better to be in some util related to dummy data, just like the other generator functions

josenriagu: okay, I will refactor it

quininez (approved):

PR #2203, opened 2024-03-08
Fix: New reflections not showing up when adding a reflection

themonster2015:
## Description

* This PR attempts to fix the issue of [new pending reflections not showing up in the reflection list after entering some text  and hitting the Reflect button](https://akasha.jetbrains.space/p/awf/issues/1300)

## Checklist


ui/hooks/src/use-pending-reflections.ts
```diff
@@ -1,28 +1,30 @@
-import { type ReactiveVar, useReactiveVar } from '@apollo/client';
+import { type ReactiveVar } from '@apollo/client';
 import type { ReflectEntryData } from '@akashaorg/typings/lib/ui';
+import { useMemo } from 'react';
 
 export const usePendingReflections = (
   pendingReflectionsReactiveVar: ReactiveVar<ReflectEntryData[]>,
 ) => {
-  const pendingReflections = useReactiveVar<ReflectEntryData[]>(pendingReflectionsReactiveVar);
+  const pendingReflections = useMemo(
```
SeverS: can you explain what you wanted to achieve here? why did you removed the userReactiveVar hook and since the pendingReflectionsReactiveVar is a function why did you store it in a useMemo?

Does this even solve the issue?

themonster2015: I did some checks and found that the pendingReflections returned from this hook was always empty even when the pendingReflectionsReactiveVar param contained the new reflection data so I thought it was not updated correctly. I attempted to fix it this way but maybe this is not the correct way. It seems to make the new reflections appear in the reflection list, like in this PR preview https://fix-new-relection-not-shown.akasha-world-framework.pages.dev/@akashaorg/app-akasha-integration/antenna

SeverS: It's not working for me in that preview. The idea is that useMemo will not trigger a rerender and you want to rerender whenever the value of the reactiveVar changes. If it works for you it's probably because something else triggers a rerender but the code added in this pr doesn't do anything to actually fix the issue.

didd (approved):

PR #2223, opened 2024-04-03
Feat: Update DS Components stories

josenriagu:
## Description

- update DS components stories

## Checklist


ui/design-system-components/src/components/Entry/EntryCard/index.tsx
```diff
@@ -148,7 +148,7 @@ const EntryCard: React.FC<EntryCardProps> = props => {
   const contentClickableStyle =
     contentClickable && !showNSFWCard ? 'cursor-pointer' : 'cursor-default';
   const menuItems: ListItem[] = [
-    ...(!isViewer
+    ...(!isViewer && flagAsLabel
```
didd: What if we don't want the label but we still want to see the menu with just the icon ? We can do
`label: flagAsLabel ?? ''`

ui/design-system-components/src/components/Entry/EntryCardRemoved/index.tsx
```diff
@@ -29,11 +29,14 @@ const EntryCardRemoved: React.FC<EntryCardRemovedProps> = props => {
...
         {props.type === 'author' ? (
           <Stack spacing="gap-y-1" customStyle="grow">
-            <Stack direction="row" spacing="gap-x-1">
-              <Text variant="button-sm">{props.message.firstPart}</Text>
-              <Text variant="footnotes2">{props.message.secondPart}</Text>
+            <Text variant="button-sm">
+              {props.message.firstPart}{' '}
+              <Text as="span" variant="footnotes2">
+                {props.message.secondPart}
+              </Text>
```
didd, line 39: Nice refactoring.

ui/design-system-components/src/components/Entry/EntryCard/index.tsx
josenriagu: That could work too. Although we do not have any such scenario yet

didd: It doesn't make sense to remove the menu item just because you don't have a label. If at some point we want icon only menu item the logic you have fails. Rather it would make sense if you check for `onEntryFlag`. If that prop isn't provided it means we don't have flagging logic yet but having empty label doesn't mean the same. It doesn't necessarily mean we don't have a flagging logic.

josenriagu: Sure, I understand your view

didd (approved):

PR #2230, opened 2024-04-09
Refactor use beams hook nsfw

themonster2015:
## Description

This PR addresses the following issues:
* Move the NSFW logic outside of the useBeams hook
* Fix the issue of disappearing subscribed content when switching between apps in My Antenna.
* Adjust the NSFW logic for My Antenna, Topic page, profile feed page
* Add the notif message when saving nsfw settings in the settings app
## Issues that will be closed

* https://akasha.jetbrains.space/p/awf/issues/1207
* https://akasha.jetbrains.space/p/awf/issues/1282
* [When Show NSFW Content is enabled user should be able to see content without previous warnings](https://akasha.jetbrains.space/p/awf/issues/1294)
*  [[BUG] User is not getting feedback when enabling Show NSFW content ](https://akasha.jetbrains.space/p/awf/issues/1292)

## Checklist


ui/lib/feed/src/components/tag-feed.tsx
```diff
@@ -42,8 +47,47 @@ const TagFeed = (props: TagFeedProps) => {
...
   const { t } = useTranslation('ui-lib-feed');
 
+  const { showNsfw } = useNsfwToggling();
+  const { data: loginData, loading: authenticating } = useGetLogin();
+  const isLoggedIn = !!loginData?.id;
+
+  let nsfwFilters;
+
+  /**
+   *  Check if the feed will be used inside the My Antenna page and
```
quininez: why is the moderation applied only to the my-antenna feed? shouldn't it also be applied to the tag feed, the one that you can navigate to by clicking on a tag?

ui/apps/profile/src/components/pages/profile-info/index.tsx
```diff
@@ -47,11 +48,20 @@ const ProfileInfoPage: React.FC<ProfileInfoPageProps> = props => {
...
   const isLoggedIn = !!loginData?.id;
   const authenticatedDID = loginData?.id;
 
+  const { showNsfw: showNsfwSetting } = useNsfwToggling();
+  const [showNSFW, setShowNSFW] = useState(false);
+
+  /* show the overlay or display the profile directly depending on the user's nsfw setting
+  and whether they are logged in or not */
+  React.useEffect(() => {
+    if (isLoggedIn && showNsfwSetting) {
```
quininez, line 60: if a logged out user has nsfw content hidden by default, and only if a user logs in can he change the nsfw settings, then maybe you can add this condition of a logged in user in the useNsfwToggling hook to simplify its usage

ui/lib/feed/src/components/tag-feed.tsx
themonster2015: I applied it to the my-antenna feed only because we decided to hide the NSFW beams when the NSFW setting is off on the My Antenna page, so I applied the filters to filter out the NSFW beams completely from the feed. However, on the tag feed page, we decided to display the nsfw beams, with an overlay, when the nsfw setting is off because we have that counter that displays how many beams there are for a specific tag, and if we hide the nsfw beams here, the number displayed won't match the number of beams a user sees on the page :)

ui/lib/feed/src/components/beam-feed.tsx
```diff
@@ -51,8 +53,47 @@ const BeamFeed = (props: BeamFeedProps) => {
     offsetTop,
   } = props;
 
+  const { showNsfw } = useNsfwToggling();
+  const { data: loginData, loading: authenticating } = useGetLogin();
+  const isLoggedIn = !!loginData?.id;
+
+  let nsfwFilters;
+
+  /**
```
didd, line 62: Is it possible to extract all this into a local function something `getNsfwFilter` ? Otherwise its getting harder to read this beam-feed file.

ui/lib/feed/src/components/beam-feed.tsx
```diff
@@ -51,8 +53,47 @@ const BeamFeed = (props: BeamFeedProps) => {
...
+          where: {
+            status: {
+              in: [AkashaBeamStreamModerationStatus.Ok, AkashaBeamStreamModerationStatus.Nsfw],
+            },
+          },
+        },
+        { where: { status: { isNull: true } } },
+      ],
+    } as AkashaBeamStreamFiltersInput;
+  }
```
didd: Is it possible to extract all this into a function something like `getNsfwFilters` ? Otherwise its getting harder to read this beam-feed file

didd (approved):

PR #2240, opened 2024-04-17
feat(): refactor login and profile hooks v1

didd:
## Description

- create user store
- use store on authentication app and sidebar widget
- fix profile placeholder ui on profile app
- fix image upload logic on profile app

## Checklist


ui/apps/profile/src/plugins/profile-plugin.ts
```diff
@@ -0,0 +1,48 @@
...
+    try {
+      const profileQuery = await this.#apolloClient.query<GetProfileByDidQuery>({
+        query: GetProfileByDidDocument,
+        variables: {
+          id: profileDid,
+        },
+      });
+
+      if (profileQuery?.errors) {
+        throw new Error(JSON.stringify(profileQuery.errors?.map(error => error.message)));
```
kenshyx: throwing and catching the error in the same block is not a good practice

typings/src/ui/store.ts
```diff
@@ -5,12 +5,12 @@ type Login = {
   checkRegistered?: boolean;
 };
 
-export interface IGetUserInfo {
+export type GetUserInfo = {
   profileDid: string;
```
kenshyx, line 9: can we rename this to `profileDID`?

didd: I have thought of doing the same change and I figured it will be easier to do it in a separate PR because we have the naming of `profileDid` outside this PR as well. I didn't want to name it `profileDid` in this PR but did it anyway to stay consistent with what we have elsewhere and make the correction on a separate PR.

kenshyx (approved):

PR #2245, opened 2024-04-19
Error handling for pasting image links

quininez:
## Checklist


ui/apps/akasha/src/extensions/image-block/image-editor-block.tsx
```diff
@@ -167,10 +162,29 @@ export const ImageEditorBlock = (
     [createBlock, retryCreate],
   );
 
+  const isImgUrl = async url => {
+    try {
+      const response = await fetch(url, { method: 'HEAD' });
+      const isImg = response?.headers?.get('Content-Type')?.startsWith('image');
+      if (isImg) return 'isImg';
```
kenshyx: why not return a boolean O_o

quininez: that was the initial approach, but it would trigger the notification message in the handle change function as well, which I didn't want

quininez: I can also return a boolean, and check for strict equality with a  boolean in the handleChange function, but I thought this is more explicit

ui/apps/akasha/src/extensions/image-block/image-editor-block.tsx
```diff
@@ -167,10 +162,28 @@ export const ImageEditorBlock = (
...
+    } catch (error) {
+      setURLNotImage(true);
+      const notifMsg = error?.message;
+      _uiEvents.current.next({
+        event: NotificationEvents.ShowNotification,
+        data: {
+          type: NotificationTypes.Error,
+          message: notifMsg,
+        },
+      });
```
didd, line 179:
```suggestion
      const notifMsg = error?.message;
      _uiEvents.current.next({
        event: NotificationEvents.ShowNotification,
        data: {
          type: NotificationTypes.Error,
          message: notifMsg,
        },
      });
      return null;
```

ui/apps/akasha/src/extensions/image-block/image-editor-block.tsx
```diff
@@ -167,10 +162,28 @@ export const ImageEditorBlock = (
...
+    }
+  };
```
didd:
```suggestion
  const isImgUrl = async url => {
      const response = await fetch(url, { method: 'HEAD' });
      const isImg = response?.headers?.get('Content-Type')?.startsWith('image');
      return isImg;
  };
```

ui/apps/akasha/src/extensions/image-block/image-editor-block.tsx
```diff
@@ -167,10 +162,28 @@ export const ImageEditorBlock = (
...
+
   const handleChange = async e => {
```
didd:
```typescript
const handleChange = async (e) => {
  const imageUrl = e.currentTarget.value;
  setImageLink(imageUrl);
  try {
    const isImg = await isImgUrl(imageUrl);
    setURLNotImage(!isImg);
    if (!isImg) {
      const notifMsg = t(`URL doesn't contain an image.`);
      _uiEvents.current.next({
        event: NotificationEvents.ShowNotification,
        data: {
          type: NotificationTypes.Error,
          message: notifMsg,
        },
      });
    }
  } catch (error) {
    setURLNotImage(true);
    const notifMsg = error.message;
    _uiEvents.current.next({
      event: NotificationEvents.ShowNotification,
      data: {
        type: NotificationTypes.Error,
        message: notifMsg,
      },
    });
  }
};
```

ui/apps/akasha/src/extensions/image-block/image-editor-block.tsx
quininez: The try catch here is for the error on the fetch, which is different from the error that can appear on the upload. Why did you suggest the change?

ui/apps/akasha/src/extensions/image-block/image-editor-block.tsx
quininez: I see now, you are catching the error here, I will test it

ui/apps/akasha/src/extensions/image-block/image-editor-block.tsx
didd: The intent is to make `isImgUrl` focus on a single concern as possible. `isImg` is ideally a pure function.  And I also noticed `isImgUrl` is only used inside `handleChange` method. 

Having the try catch and sending the notification is fine in my opinion then it will no longer be pure function. And in this case perhaps we return null inside the catch block. And inside handleChange function we can check it using `Object.is(isImg, null)`. On the other hand, setting the state inside this function is making it take care of more concerns than it should.

didd (approved):

PR #2254, opened 2024-04-26
use-reflection hook - switch to ReflectionStream

themonster2015:
## Checklist


ui/apps/akasha/src/components/pages/entry-page/beam-page.tsx
```diff
@@ -106,48 +106,41 @@ const BeamPage: React.FC<BeamPageProps> = props => {
...
-            </>
-          )}
+                logger={logger}
+              >
+                <>
+                  <Divider />
+                  <EditableReflectionResolver
+                    reflectID={itemData.node.reflectionID}
+                    beamID={beamId}
+                  />
```
didd, line 125: Need to expose the `reflectToId` for this component and set it as `beamId` otherwise it will not be correct. Can you check this closely ?

ui/hooks/src/use-reflections.ts
```diff
@@ -88,26 +92,24 @@ export const useReflections = (props: UseReflectionProps) => {
...
     return vars;
-  }, [entityId, sorting, filters]);
+  }, [sorting, filters]);
 
-  const [fetchReflections, reflectionsQuery] = getterHook({
+  const [fetchReflections, reflectionsQuery] = useGetReflectionStreamLazyQuery({
     variables: {
       ...mergedVars,
-      first: overscan,
+      last: overscan,
```
didd: Is this actually `last` or indeed `first` ? Did you test it ?

ui/lib/feed/src/components/editable-reflection/editable-reflection-resolver.tsx
```diff
@@ -1,50 +1,54 @@
...
-            const reflectionId = reflectionQuery.data.node.id;
+          if (hasOwn(reflectionReq.data?.node, 'id')) {
+            const reflectionId = reflectionReq.data.node.id;
             return getRoutingPlugin().navigateTo({
               appName: '@akashaorg/app-akasha-integration',
-              getNavigationUrl: navRoutes => `${navRoutes.Reflect}/${reflectionId}`,
+              getNavigationUrl: navRoutes => `${navRoutes.Reflection}/${reflectionId}`,
             });
           }
         }}
```
didd: We can expose this onContentClick as a prop and let the parent handle it. That way we won't need to use navigateTo from the routing plugin but can use the navigate from TanStack router which has TypeScript support.

ui/lib/feed/src/components/editable-reflection/editable-reflection-resolver.tsx
```diff
@@ -1,50 +1,54 @@
...
-            const reflectionId = reflectionQuery.data.node.id;
+          if (hasOwn(reflectionReq.data?.node, 'id')) {
+            const reflectionId = reflectionReq.data.node.id;
             getRoutingPlugin().navigateTo({
               appName: '@akashaorg/app-akasha-integration',
               getNavigationUrl: navRoutes =>
-                `${navRoutes.Reflect}/${reflectionId}${navRoutes.Reflect}`,
+                `${navRoutes.Reflection}/${reflectionId}${navRoutes.Reflect}`,
             });
           }
```
didd: Same comment as above

ui/lib/feed/src/components/editable-reflection/index.tsx
```diff
@@ -69,6 +69,8 @@ const EditableReflection: React.FC<ReflectCardProps & { reflectToId: string }> =
     onCompleted: async () => {
       setEdit(false);
       setNewContent(null);
+      //reset the state modified inside Editor component when done
+      setIsReflecting(true);
```
didd: Is this meant to be `false` if you are resetting ?

themonster2015: No it means `true` because we need to reset the state that would be set as `false` inside the child component and the original state of `isReflecting` is true

didd: To make the intent clear what if you create another component inside the same folder something like ReflectionEditorRenderer or something simmilar that wraps the reflection editor along with the state to show and hide the editor. And also rename the local state to showEditor rather than `isReflecting`. That way you don't need to reset it here and the intent will be more clear.

themonster2015: You created this component a while ago Didd :)). I will try your suggestion to make it clearer

didd: Yes but the initial intent has been changed ever since :) Thanks.

didd: Yes I created the component initially. But it has evolved over time and what I saw on this PR made it clear the naming for the local state and the reset here doesn't make sense :) Thanks.

ui/hooks/src/use-reflections.ts
themonster2015: I tested both and if I use a high overscan number (120), it doesn't really make any difference if it's `first` or `last` and it still displays fine. But if I use a low `overscan` number, it will shift the reflections no matter whether it's `first` or `last`.

SeverS: this is tricky as the overscan defines how many items to fetch at once, having it set to a high number will only hide the problem (aka overscan=120 for a small list it basically means fetch/refetch everything) and it will bite us in the future.

themonster2015: I see that the `estimatedHeight` is also set to 120 when being passed to the Virtual List. Does this number mean something related to the estimated number of items?

SeverS: no, that is the estimated height of a card - this estimation is used before measuring the actual height of the card

themonster2015: Thanks for explaining that :). Now I'm testing `overscan` with 25

PR #2263, opened 2024-05-08
Patch: cookie widget

josenriagu:
## Description

- treat edge case when widget area overscrolls

## Checklist


ui/widgets/layout/src/components/layout-widget.tsx
```diff
@@ -246,7 +246,7 @@ const Layout: React.FC<unknown> = () => {
               />
             </Stack>
 
-            <Stack customStyle="fixed bottom-2 mx-2 lg:(static my-4)">
+            <Stack customStyle="fixed bottom-2 mx-2 lg:(w-[21rem] mb-2)">
```
didd: Not sure if giving this a fixed width is good idea may be max width on the entire widgets container works better. And in that case the position need to change to absolute and give the widget container position relative. This needs some thought for a better approach. Curious to see what others think here.

themonster2015 (changes requested):
Screenshot 2024-05-10 at 04 29 42

It looks like with this approach, when scrolling down a feed, the latest topics and profile widgets disappear leaving a big empty space

extensions/widgets/layout/src/components/layout-widget.tsx
```diff
@@ -234,7 +234,7 @@ const Layout: React.FC<unknown> = () => {
             </Stack>
           </Stack>
 
-          <Stack align="center" justify="between" customStyle="sticky top-0 h-screen">
+          <Stack align="center" justify="between" customStyle="sticky top-0 h-screen relative">
```
didd: relative and sticky are mixed here, I understand the relative part is a left over from your local test. And that should fix Yen's observation.

themonster2015 (approved):

ui/widgets/layout/src/components/layout-widget.tsx
themonster2015: Just saw the comment. I agree that giving a fixed width may not be a good idea

PR #2320, opened 2024-06-19
Feat: Extensions Developer Mode

josenriagu:
## Description

- add dev mode toggle option to settings

## Checklist


libs/hooks/src/use-dev-mode.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @example useDevMode hook
+ * ```typescript
+ * const { devMode, propagateDevMode } = useDevMode();
+ * ```
+ * @returns `{ devMode, propagateDevMode }` The current devMode value and a function to set the devMode preference.
+ * The `propagateDevMode` function takes just a boolean parameter indicating whether or not you are enabling dev mode.
+ **/
+export const useDevMode = (): TUseDevModeReturn => {
+  // get the dev mode preference, if any, from local storage
+  const isDevMode = window.localStorage.getItem('DevMode');
```
kenshyx: - use a const for the key
- keys should be lowercase
- use namespace separator `:` to prepend also the app-name in the key

libs/hooks/src/use-dev-mode.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  const [devMode, setDevMode] = useState<boolean>(isDevMode === 'true');
+
+  const propagateDevMode = useCallback((_isDevMode: boolean) => {
```
kenshyx: there is no need for a react hook, the page needs to full refresh after setting the dev mode

extensions/apps/extensions/src/index.tsx
```diff
@@ -18,47 +18,63 @@ import { Akasha } from '@akashaorg/design-system-core/lib/components/Icon/akasha
...
-  i18nNamespace: ['app-extensions'],
-  routes: {
-    ...routes,
-  },
-  title: 'AKASHA Core Extensions',
-  menuItems: {
-    label: 'Extensions',
-    type: MenuItemType.App,
+export const register: (opts: IntegrationRegistrationOptions) => IAppConfig = opts => {
+  const isDevModeEnabled = window.localStorage.getItem('DevMode') === 'true';
```
kenshyx: create a function that handles all this business logic

kenshyx (approved):

PR #2322, opened 2024-06-20
Create extension page

themonster2015:
## Checklist


libs/design-system-components/src/components/AppCreationForm/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  cancelButton,
+  createButton,
+  errorMessage,
+}) => {
+  const {
+    control,
+    setError,
+    clearErrors,
+    handleSubmit,
+    formState: { errors, isDirty, isValid },
```
didd: `isDirty` and `isValid` aren't deterministic as explained on this issue https://akasha.jetbrains.space/p/awf/issues/1024 ... there is an istance where their computation isn't accurate. For example while you have errors `isValid` becomes true. It's much safer to use `dirtyFields` and `errors` objects to compute the flags. 

```
formState: { errors, dirtyFields, errors },

....
const isDirty = !!Object.keys(dirtyFields).length;
const isValid = !Object.keys(errors).length;
```

And the `onSubmit` needs to be changed as well.

You can see a sample of how this is used on `EditProfile` component.

themonster2015: I tried using it the way you suggested but it incorrectly enabled the 'Create' button as soon as one field changes even though the remaining required fields are still empty/not touched. In my case, I wanted to disable the button completely till the required fields have been completed. I will test more and see why.

themonster2015: Figured it out. Need to use a different condition for isFormDirty :)

libs/design-system-components/src/components/AppCreationForm/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+  };
+
+  return (
+    <form onSubmit={onSave} className={tw(apply`h-full`)}>
+      <Stack direction="column" spacing="gap-y-4">
+        <Stack padding="px-4 pb-16" spacing="gap-y-4">
+          <Controller
+            control={control}
+            name={FieldName.extensionType}
+            render={({ field: { name, value = extensionTypes[0], onChange, ref } }) => (
```
didd: You have default value set, setting value to `extensionTypes[0]` shouldn't be necessary.

themonster2015: thanks for noticing it Didd.

didd (approved):

PR #2330, opened 2024-06-26
Fix: remove beams

josenriagu:
## Description

- update List element click handler
- remove beams in succession

## Checklist


didd (dismissed):

libs/design-system-core/src/components/List/list-element.tsx
```diff
@@ -9,15 +9,14 @@ export type ListItemProps = { customStyle?: string } & ListItem;
 
 const ListElement: React.FC<ListItemProps> = props => {
   const { label, icon, color, disabled, customStyle = '', onClick, ...rest } = props;
+
+  const handleButtonClick = (e: React.SyntheticEvent) => {
+    e.stopPropagation();
```
didd: On a second thought, a component needs to use `stopPropagation()` implies a component further up the chain will break something. It means child component know about the parent component which isn't a good design. In this case need to fix the real issue rather than introducing a workaround. What do you think ?

josenriagu: I added this because the list items were used in the menu which is popover like and could propagate clicks. Are you suggesting we define this stopPropagation() to the handlers themselves?

didd: Yes it makes more sense to use stopPropagation() `onEntryFlag` and `onEntryRemove` inside EntryCard component.

josenriagu: alright

SeverS (approved):

PR #2342, opened 2024-07-11
My extensions Infinite list

quininez:
## Description
* add infinite scroll and filter functionality to my extensions page
* fix mini profile widget cover image source

## Checklist


extensions/apps/extensions/src/components/pages/my-extensions/extension-element.tsx
```diff
@@ -0,0 +1,251 @@
...
+      case AkashaAppApplicationType.Widget:
+        return <Widget />;
+    }
+  };
+
+  const getExtensionStatus = () => {
+    if (appStreamData?.edges?.length === 0) {
+      return ExtensionStatus.Draft;
+    }
+    if (appStreamData?.edges[0]) {
```
kenshyx: refactor this to use a switch statement with the default return type being `Draft`

extensions/apps/extensions/src/components/pages/my-extensions/extension-element.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const { navigateToModal } = useRootComponentProps();
+
+  const { data: appStreamReq } = useGetAppsStreamQuery({
+    variables: {
+      indexer: sdk.current.services.gql.indexingDID,
+      first: 1,
+      filters: {
+        where: {
+          applicationID: {
+            equalTo: extensionData.id,
```
kenshyx, line 67: ensure that the id is not empty

quininez: on line 72 there is the option passed to skip the query if id is nullish

libs/design-system-core/src/utils/string-utils.ts
```diff
@@ -1,3 +1,11 @@
+export const capitalize = (str?: string) => {
```
didd: You can use capitalize utility function from lodash.

extensions/apps/extensions/src/components/pages/my-extensions/extension-element.tsx
kenshyx: a string like `" "` would pass the validation, we need more strict checks for filters

extensions/apps/extensions/src/components/pages/my-extensions/extension-element.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const showElement = () => {
+    if (filter.id === '0') {
+      return true;
+    }
+    return filter.title === getExtensionStatus();
+  };
+
+  return (
+    <>
+      {showElement() && (
```
didd, line 186:
```suggestion
      {filter.id === '0' || filter.title === getExtensionStatus() && (
```

extensions/apps/extensions/src/components/pages/my-extensions/extension-element.tsx
```diff
@@ -0,0 +1,251 @@
...
+  };
+
+  const handleExtensionRemove = () => {
+    navigateToModal({
+      name: `remove-extension-confirmation`,
+      extensionId: extensionData.id,
+    });
+  };
+
+  const menuItems = (extensionStatus: string) => {
```
didd:
```suggestion
  const menuItems = (extensionStatus: string) : MenuProps['items'] | [] => {
```
That instead of type assertions on each case.

extensions/apps/extensions/src/components/pages/my-extensions/extension-element.tsx
quininez: I think the function is more explicit in this case

didd (approved):

PR #2343, opened 2024-07-12
Add license to extension creation form

quininez:
## Checklist


libs/design-system-components/src/components/AppCreationForm/index.tsx
```diff
@@ -64,11 +81,23 @@ const AppCreationForm: React.FC<AppCreationFormProps> = ({
     }
   }, [errorMessage, errors]);
 
+  const extensionLicenseValue = watch(FieldName.extensionLicense);
```
didd: I think it's safer to use `useWatch` to mitigate any performance issues. But if you think it's fine I don't have strong opinion on this as it's used right now only on one field.

didd (approved):

PR #2344, opened 2024-07-16
Fix: Profile interests

josenriagu:
## Description

- update profile interest click handler
- add snackbar notification to topic handlers
- fix breaking stories

## Checklist


extensions/widgets/trending/src/components/cards/latest-topics.tsx
```diff
@@ -34,6 +38,15 @@ export type LatestTopicsProps = {
   refetchTagSubscriptions: () => void;
 };
 
+/**
+ * This utility method compares two arrays of tags and returns the difference,
+ * @param arr1 - first array to be compared
+ * @param arr2 - second array to be compared
+ * @returns an array of strings
+ */
+const getTagsDiff = (arr1: string[], arr2: string[]) =>
```
kenshyx: what if the lists are sorted different? this doesn't look it does what it is expected

extensions/widgets/trending/src/components/cards/latest-topics.tsx
```diff
@@ -1,17 +1,26 @@
...
 import { TopicRow } from './topic-row';
-import isEqual from 'lodash/isEqual';
-import difference from 'lodash/difference';
-import pullAll from 'lodash/pullAll';
+
+declare global {
+  interface Set<T> {
+    symmetricDifference: (value: Set<T>) => Set<T>;
+  }
+}
```
josenriagu: @kenshyx is it okay, if we have this global defined here?

extensions/widgets/trending/src/components/cards/latest-topics.tsx
```diff
@@ -34,6 +43,19 @@ export type LatestTopicsProps = {
...
+ * This utility method compares two arrays of tags and returns the difference,
+ * @param arr1 - first array to be compared
+ * @param arr2 - second array to be compared
+ * @returns an array of strings
+ */
+const getTagsDiff = (arr1: string[], arr2: string[]) => {
+  const setA = new Set(arr1);
+  const setB = new Set(arr2);
+
+  return Array.from(setA.symmetricDifference(setB));
```
josenriagu, line 50: @kenshyx used like so, here

extensions/widgets/trending/src/components/cards/latest-topics.tsx
kenshyx: why would you do something like this? this is a type that should come from the typescript lib

kenshyx: merge with next, I've updated the tsconfig for you;
please remove this global overwrite

libs/typings/src/ui/ui-events.ts
```diff
@@ -56,7 +56,6 @@ export const enum NotificationEvents {
  **/
 export const enum NotificationTypes {
   Info = 'info',
-  Alert = 'alert',
```
kenshyx, line 59: why `Alert` type had to be removed?

josenriagu: It was replaced with Info, per the updated design
https://www.figma.com/design/WbIHLctkDhGJjLKLhZcjCf/AKASHA-WORLD-DESIGN-SYSTEM?node-id=13-2121&t=CDxoD4c4899D6Wrj-0

josenriagu: We now have only 4 notif types: Info, Caution, Success, Error

kenshyx (dismissed):

didd (approved):
The PR doesn't address race condition while subscribing and eventually notifying users the topics they subscribed to. I have asked Jose to create a task to handle this.

kenshyx (approved):

PR #2349, opened 2024-07-24
Extension edit form

quininez:
## Description
* multi step form wizard for editing extensions
* saves form data on session storage between steps
* on submit the data is updated on local storage
* has fields for logoImage, coverImage, name, displayName, description, nsfw, links, contributors, gallery, contactInfo
* update extension creation form to save on local storage

## Checklist


extensions/apps/extensions/src/components/pages/extension-creation-page.tsx
```diff
@@ -1,29 +1,38 @@
...
 import { NotificationEvents, NotificationTypes } from '@akashaorg/typings/lib/ui';
 import Card from '@akashaorg/design-system-core/lib/components/Card';
 import Stack from '@akashaorg/design-system-core/lib/components/Stack';
 import Text from '@akashaorg/design-system-core/lib/components/Text';
 import Divider from '@akashaorg/design-system-core/lib/components/Divider';
-import AppCreationForm, {
-  FieldName,
-} from '@akashaorg/design-system-components/lib/components/AppCreationForm';
+import AppCreationForm from '@akashaorg/design-system-components/lib/components/AppCreationForm';
+import { NotConnnected } from '../not-connected';
```
SeverS: You can use ErrorLoader component from the ds-core

quininez: yes, I will change it, the error loader was updated in Jose's PR

libs/design-system-components/src/components/ExtensionEditStep2Form/DocumentationLinks/link-element.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
```
SeverS, line 1: shouldn't this component be in the ds-core? and also made more generic

josenriagu: We already have the Link component in DS Core

libs/design-system-components/src/components/ExtensionEditStep2Form/NSFW/index.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
```
SeverS: This is duplicate of ds-components/EditProfile/NSFW

quininez: it's the same functionality, but the control passed from react-hook-form is different here

libs/design-system-components/src/components/ExtensionEditStep2Form/DocumentationLinks/link-element.tsx
quininez: yes, this could be improved, to be transformed in a more generic component, although I would create a separate task for it

quininez: @josenriagu the link component we have in DS performs a different role

libs/design-system-components/src/components/ExtensionEditStep2Form/NSFW/index.tsx
SeverS: you mean the typings for the control object because the control is passed as a prop

quininez: yes, the typings, do you think it would make sense to have more generic typings for it and reuse the component?

libs/typings/src/ui/extensions.ts
```diff
@@ -16,6 +22,33 @@ export enum ExtensionStatus {
   Published = 'Published',
 }
 
+/**
+ * Type defining extension image types
+ **/
+export type ExtensionImageType = 'logo-image' | 'cover-image';
+
+export type Extension = {
```
SeverS: Can't we reuse the AkashaAppInterface here? no need to redefine all the fields, just add the extra ones

SeverS:
Don't forget to add issues for the components that can be moved to ds-core (link-element, nsfw, search-bar)

quininez:
nsfw I moved to a standalone component in ds-components already and I created issues for the other 2

SeverS:
I saw that, but it can be moved to the ds-core because it's generic enough to be a FormCheckbox component

SeverS (dismissed):

didd (approved):

josenriagu (approved):

PR #2350, opened 2024-07-25
Fix: My antenna

josenriagu:
## Description

- update tag filters so that matching beams can be correctly fetched

## Checklist


libs/feed/src/components/tag-feed.tsx
```diff
@@ -85,6 +81,11 @@ const TagFeed = (props: TagFeedProps) => {
     );
   }
 
+  // @TODO: replace with cache policy
```
SeverS: are you going to address this todo in this pr?

libs/feed/src/components/tag-feed.tsx
```diff
@@ -85,6 +81,11 @@ const TagFeed = (props: TagFeedProps) => {
...
+  const filteredBeams = beams.filter(
```
SeverS: can you wrap this into a useMemo so it not iterate on every re-render?

libs/feed/src/components/tag-feed.tsx
josenriagu: No, it will be a different task

SeverS (approved):

didd (approved):

PR #2355, opened 2024-07-29
chore(): refactor user store

didd:
## Description

- refactor user store
   - rename store 
   - fix infinite authentication loader from being displayed when there is no active session 
   - create fallback profile function for profile plugin absence
 - fill editor space when a beam is removed

## Checklist


extensions/apps/antenna/src/components/pages/entry-page/beam-section.tsx
```diff
@@ -25,10 +26,15 @@ const BeamSection: React.FC<BeamSectionProps> = props => {
...
 
   return (
-    <Stack dataTestId="beam-section" spacing="gap-y-2" customStyle="grow mb-2">
-      <Stack customStyle="grow">
+    <Stack
+      dataTestId="beam-section"
+      spacing="gap-y-2"
+      customStyle={`grow ${activeBeam ? 'mb-2' : ''}`}
+    >
+      <Card customStyle="grow" type="plain">
```
josenriagu, line 37: any preference for switching to Card type of 'plain' instead of the Stack?

didd: yes, there is no need to apply any style except flex grow

josenriagu (dismissed):

libs/hooks/src/store/use-akasha-store.ts
```diff
@@ -1,11 +1,17 @@
...
+import { AkashaProfile } from '@akashaorg/typings/lib/ui';
 
 export function useAkashaStore() {
   const { plugins } = useRootComponentProps();
-  const userStore = plugins['@akashaorg/app-profile']?.profile
-    ?.userStore as IUserStore<AkashaProfile>;
-  const data = useSyncExternalStore(userStore?.subscribe, userStore?.getSnapshot);
-  return { data, userStore };
+  const authenticationStore = AuthenticationStore.getInstance<AkashaProfile>(
+    plugins['@akashaorg/app-profile']?.profile?.getProfileInfo || getProfileInfo,
```
SeverS: I would rename the fallback `getProfileInfo` to `getAkashaProfileInfo`. will make it more clear

SeverS: I'm still not very happy that the authenticationStore works only with `plugins['@akashaorg/app-profile']` plugin. I'm thinking more and more on splitting this into 2 stores without any deps between them. A store that requires a DID and implements a profile plugin. Can you create an issue with this? We can work on it afterwards

libs/typings/src/ui/store.ts
```diff
@@ -21,12 +21,12 @@ export interface IUserState<T> {
...
 
 /**
- * Interface defining a user state store
+ * Interface defining authentication state store
  */
-export interface IUserStore<T> {
+export interface IAuthenticationStore<T> {
   login({ provider, checkRegistered }: Login): void;
   logout(): void;
   restoreSession(): void;
```
SeverS: when should `restoreSession` be called? Since the store is a sigleton doesn't make sense to be called internally?

didd: it made sense to call the restore session when the singleton instance is created which happens during page refresh

libs/hooks/src/store/use-akasha-store.ts
didd: You mean one store just for getting the plugin that has function to fetch profile info and the other one the authentication store ?

libs/typings/src/ui/store.ts
didd: do you see others cases where it should be called ?

libs/hooks/src/store/use-akasha-store.ts
SeverS: yes, well a store that implements only one method does not make much sense now, right? This is why it's not a high priority right now. But in the near future we can  include the methods to handle all CRUD operations via the profile plugin.

libs/typings/src/ui/store.ts
SeverS: Idk. this is why i've asked. But I can see that you are calling it in the constructor. then I would make it private

libs/hooks/src/store/use-akasha-store.ts
didd: done, I have created the issue on linear

SeverS (dismissed):

SeverS (approved):

josenriagu (approved):

PR #2360, opened 2024-08-02
Feat: DS stepper component

josenriagu:
## Description

- create stepper component for stepped actions
- add and update stories

## Checklist


quininez:
From the design I can see that the component has 3 states for each step: a grey indicator when the step is not reached, a filled accent circle for a completed step and an empty circle for the current step. From what I can see in the implementation the component now only has 2 states.

libs/design-system-components/src/stories/dscomponents/stepper.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import Stepper, { StepperProps } from '../../components/Stepper';
+
+Stepper.displayName = 'Spinner';
```
SeverS: stepper's display name is spinner?

josenriagu: fixing this now

josenriagu:
> From the design I can see that the component has 3 states for each step: a grey indicator when the step is not reached, a filled accent circle for a completed step and an empty circle for the current step. From what I can see in the implementation the component now only has 2 states.

Screenshot 2024-08-05 at 09 03 29
This is what it looks like on the story.

quininez (approved):

SeverS (approved):

PR #2371, opened 2024-08-15
Fix: App specific widgets

josenriagu:
## Description

- update app-loader logic to render `app-specific widgets`

## Checklist


libs/app-loader/src/index.ts
```diff
@@ -408,7 +407,10 @@ export default class AppLoader {
       if (!conf.loadingFn || typeof conf.loadingFn !== 'function') continue;
 
       const extensionData = this.extensionData.find(m => m.name === name);
-      if (extensionData.applicationType !== AkashaAppApplicationType.App) continue;
+
+      // exclude only widgets that are always active (ie activeWhen is not explicitly defined)
+      if (extensionData.applicationType !== AkashaAppApplicationType.App && !conf.activeWhen)
```
SeverS: widgets (of any kind) should not register into the single-spa here. AppLoader only handles APPS. What you are doing here basically you will leak any kind of extensionType that doesn't have activewhen property set. You get double sidebar, topbar, etc. because it will register here and then they will also register into the widgetStore and be mounted from there also. You need to check the widgetStore along with the Widget component in libs/extensions/src/react/widget.tsx. The problem should be there

josenriagu: okay, I will take a look

SeverS (approved):

kenshyx (approved):

PR #2384, opened 2024-08-29
fix(): cors issue on create image utility function

didd:
## Description

- fix cors issue on create image utility function

## Checklist

### DIFF
--- a/libs/design-system-core/src/utils/get-cropped-image.ts
+++ b/libs/design-system-core/src/utils/get-cropped-image.ts
@@ -26,11 +26,22 @@ const rotateSize = (width: number, height: number, rotation: number) => {
  */
 const createImage = (url: string): Promise<HTMLImageElement> =>
   new Promise((resolve, reject) => {
-    const image = new Image();
-    image.addEventListener('load', () => resolve(image));
-    image.addEventListener('error', error => reject(error));
-    image.setAttribute('crossOrigin', 'anonymous');
-    image.src = url;
+    fetch(url)
+      .then(response => response.blob())
+      .then(blob => {
+        const image = new Image();
+        const imageObjectUrl = URL.createObjectURL(blob);
+        image.src = imageObjectUrl;
+        image.onload = () => {
+          URL.revokeObjectURL(imageObjectUrl);
+          resolve(image);
+        };
+        image.onerror = error => {
+          URL.revokeObjectURL(imageObjectUrl);
+          reject(error);
+        };
+      })
+      .catch(error => reject(error));
   });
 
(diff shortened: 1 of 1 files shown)


libs/design-system-core/src/utils/get-cropped-image.ts
```diff
@@ -26,11 +26,22 @@ const rotateSize = (width: number, height: number, rotation: number) => {
  */
 const createImage = (url: string): Promise<HTMLImageElement> =>
   new Promise((resolve, reject) => {
```
SeverS, line 28: no need to wrap into a new promise, fetch is already returning a promise. you can also use async/await.
how is this resolving the cors issue?

didd: - the promise is there to wait for the image to load, then to resolve with the image otherwise if during loading there is an error to reject with the error 
- the previous approach required to set crossOrigin attribute on the image element and the options where limited which resulted in the opaque response because there was a mismatch in how the server expected to handle the response and how the request was made ... but now when using fetch the default mode is to use cors and this time there was no mismatch

SeverS (approved):

libs/design-system-core/src/utils/get-cropped-image.ts
kenshyx: cors should be mandatory for loading content from 3rd party servers

kenshyx: the proposed solution could be replaced by removing just 1LOC `image.setAttribute('crossOrigin', 'anonymous');`

kenshyx: but if this approach has better results then we should observe for possible issues on next

kenshyx (approved):

libs/design-system-core/src/utils/get-cropped-image.ts
kenshyx: you could even return just the `ObjectURL` since this one gets transformed back again from the image object

kenshyx: and `revokeObjectURL` inside the callback function from canvas.toBlob

didd: actually removing that one line will throw an error when trying to access the canvas image data ... it gives an error saying canvas has been tainted by cross-origin data but the fetch approach is clean imo we can observe the behavior on next ... yes I could return the object but I need the image object to access the dimensions and such info

PR #2386, opened 2024-08-30
chore(): bug fixes v2

didd:
## Description

- offset calculation on virtual list component
- apply line break on text editor when the text would overflow its content box
- set maximum characters limit to profile name as 50
-  prevent pending reflection card from appearing when there is error during publishing
- set maximum encoded length for reflect editor as 3000 rather than the default 6000

## Checklist

### DIFF
--- a/extensions/apps/antenna/src/components/reflect-editor/index.tsx
+++ b/extensions/apps/antenna/src/components/reflect-editor/index.tsx
@@ -53,13 +53,16 @@ const ReflectEditor: React.FC<ReflectEditorProps> = props => {
   const sdk = getSDK();
   const isReflectOfReflection = reflectToId !== beamId;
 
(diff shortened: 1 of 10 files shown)


josenriagu (approved):

extensions/apps/antenna/src/components/reflect-editor/index.tsx
```diff
@@ -85,32 +88,42 @@ const ReflectEditor: React.FC<ReflectEditorProps> = props => {
...
-        {
-          type: 'paragraph',
-          children: [
-            {
-              text: editorValueForTest,
-            },
-          ],
-        },
-      ]);
+      setEditorState(
```
SeverS, line 97: I'm not convinced that this is the way to do it. Maybe it was better for this component to allow an intialState of slate content type (passed as props). Please create an issue with this to be addressed later.

SeverS (approved):

PR #2402, opened 2024-09-13
Chore: UI fixes

josenriagu:
## Description

- update Tooltip component UI
- increase spacing between blocks on beam card
- update Auth screen UI
- change icon size in profile settings

## Checklist


quininez (dismissed):

didd:
image
 It seems the arrow's color on the tooltip is different from the content background.

didd (dismissed):

quininez:
in the design it looks like the shadow follows also the arrow of the tooltip on light theme, and on dark theme there is no shadow at all, which is different from the implementation.  Right now in the implementation on dark theme the arrow looks weird and breaks the flow of the component.
https://www.figma.com/design/WbIHLctkDhGJjLKLhZcjCf/AKASHA-WORLD-DESIGN-SYSTEM?node-id=923-6497&node-type=canvas&t=zoCKN9ozKBtP4FUN-0

maybe @Basmatiii can take a look at this and clarify

didd (dismissed):

quininez (commented):
I checked the branch preview and it doesn't  seem like the shadow is following the arrow of the tooltip

SeverS:
Design approved! Let's goooo!

libs/design-system-core/src/components/Button/button-icon.tsx
```diff
@@ -38,7 +38,7 @@ export const ButtonIcon: React.FC<
     <Icon
       icon={icon}
       solid={solid}
-      size={size}
+      size={size === 'xs' ? 'sm' : size}
```
SeverS, line 41: hmm, i wonder why is acrobatics required. what if we pass the iconSize as a separate prop?

josenriagu: this is for the icon only button when the passed size if very small 'xs'. rather than specifying a new set of width/height for that singular use case.

quininez:
I checked again, I had a cached version, looks good now

quininez (approved):

SeverS (approved):

PR #2411, opened 2024-09-18
Install Progress page

SeverS:
## Checklist


package.json
```diff
@@ -4,9 +4,9 @@
   "scripts": {
     "clean": "tsc --build --clean tsconfig.json",
     "clean:all": "yarn clean && nx reset",
-    "prepare": "nx run-many --target=prepare --all --verbose",
-    "build:libs": "nx run-many -t build --projects=tag:type:lib --verbose",
-    "build:apps": "nx run-many -t build --projects=tag:scope:extension --verbose",
+    "prepare": "nx run-many --target=prepare --all --verbose --skip-nx-cache",
```
kenshyx: skipping the nx cache defeats the whole purpose

worlds/akasha.world/webpack.config.cjs
```diff
@@ -13,7 +13,7 @@ module.exports = Object.assign(baseConfig, {
   context: path.resolve(__dirname),
   entry: {
     main: './src/index.ts',
-    'setup.twind': './src/setup.twind.ts'
+    'setup.twind': './src/setup.twind.ts',
```
kenshyx: I'm not sure if the trailling comma is ok in cjs files

package.json
kenshyx: if you want to skip the nx cache locally on your machine then you could set the env value `NX_SKIP_NX_CACHE=true`

kenshyx: without adding this flag by default to all the build commands

SeverS: oops, the package.json shouldn't be pushed, everyone is handling this on their own. reverting

worlds/akasha.world/webpack.config.cjs
SeverS: is just my editor fixing things... reverting this

extensions/apps/extensions/src/components/pages/install-extension/install-extension-page.tsx
```diff
@@ -161,51 +214,63 @@ export const InstallExtensionPage = ({ appId }: { appId: string }) => {
...
   }, [data, isInstalled, reportedError, reportedStatus, t]);
 
-  const action = useMemo(() => {
+  const successLabel = useMemo(() => {
     if (isInstalled) {
-      return {
-        label: t('Open the app'),
+      return t('Finished');
+    }
+    if (shouldRegisterResources && 1 < 0 /*successfullySigned*/) {
```
didd, line 245: there is something that wasn't updated here otherwise this will always be false

SeverS: we don't have the new resource sign flow, will be updated when we'll have it

didd (approved):

quininez (approved):

PR #2418, opened 2024-09-25
Fix: Nsfw cards click on Global Antenna

josenriagu:
## Description

- make nsfw cards clickable when set to show on the global antena

## Checklist


libs/feed/src/components/beam-content-resolver.tsx
```diff
@@ -13,7 +13,7 @@ export type BeamContentResolverProps = {
 
 const BeamContentResolver: React.FC<BeamContentResolverProps> = ({
   beamId,
-  showNSFWCard,
+  showNSFWCard = false,
```
SeverS, line 21: Can you explain a bit how is this flag solving the cards click and why isn't correctly passed from the parent?

josenriagu: it has to be specified so as to not show nsfw contents on the feed and also used to determine when to attach the content click handler on the entry card. the prop is optional and this component is used in multiple other pages like my antenna, tag feed page etc, rather than set  it on all other places, I thought to add a default value as a guard. what do you think?

SeverS (approved):

kenshyx (approved):

PR #2422, opened 2024-09-27
Improve extension form validation

quininez:
## Description
* enforce unique extension name on creation and edit extension forms
* validate links to ensure href is an URL

## Checklist


libs/design-system-components/src/components/ExtensionEditStep2Form/index.tsx
```diff
@@ -198,4 +192,14 @@ const schema = z.object({
     .optional()
     .transform(e => (e === '' ? undefined : e)),
   nsfw: z.boolean().optional(),
+  links: z.array(
```
SeverS: links are max 10 - according to the model

quininez: I do the check somewhere else, but I will add it also here

SeverS (approved):

josenriagu (approved):

PR #2425, opened 2024-09-30
fix(): add optional chaining everywhere a profile object is used

didd:
## Description

- add optional chaining everywhere a profile object is used

## Checklist

### DIFF
--- a/libs/design-system-components/src/components/ExtensionEditStep3Form/Collaborators/index.tsx
+++ b/libs/design-system-components/src/components/ExtensionEditStep3Form/Collaborators/index.tsx
@@ -62,8 +62,8 @@ export const Collaborators: React.FC<CollaboratorsProps> = ({
     }
   };
 
-  const handleRemoveContributor = (profile?: AkashaProfile) => {
-    setAddedContributors(prev => prev.filter(addedProfile => profile?.id !== addedProfile?.id));
+  const handleRemoveContributor = (profile: AkashaProfile) => {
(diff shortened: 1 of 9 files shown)


libs/design-system-components/src/components/ExtensionEditStep3Form/Collaborators/index.tsx
```diff
@@ -63,7 +63,7 @@ export const Collaborators: React.FC<CollaboratorsProps> = ({
   };
 
   const handleRemoveContributor = (profile: AkashaProfile) => {
-    setAddedContributors(prev => prev.filter(addedProfile => profile.id !== addedProfile?.id));
+    setAddedContributors(prev => prev.filter(addedProfile => profile?.id !== addedProfile?.id));
```
kenshyx: this one is not optional

kenshyx: or if it's optional, couldn't this lead to false positive?

didd: I made it optional on the param. From what I see in the code there is no guarantee the profile param will not be null.

didd: I thought more about this, yeah I have scraped the change. It doesn't actually make sense to make the param optional here.

kenshyx (approved):

PR #2427, opened 2024-10-01
Populate ExtensionPage with real data

SeverS:
## Checklist


extensions/apps/extensions/src/components/pages/info-page/AppCoverImage/index.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { useMemo } from 'react';
```
quininez, line 1: this can be used also in the edit extension form, I can add it there when I work on edit published extension, but for now maybe it fits better in design system core, just as we have AppAvatar

SeverS: My thought was to move it to ds-core only if there are multiple apps that needs it. the edit form is also in the same app (extensions) so it can import it from there. we can create a components folder inside the app for shared components

quininez: ok, so if we run into the case where other components outside the extensions app need to use it, then we move it to ds-core, otherwise it stays here

quininez (approved):

didd (approved):

PR #2428, opened 2024-10-01
chore(): show list of installed apps on installed extensions page

didd:
## Description

-  show list of installed apps on installed extensions page

## Checklist


kenshyx (approved):

libs/design-system-core/src/components/ErrorLoader/error-card.tsx
```diff
@@ -31,12 +31,20 @@ const ErrorCard: React.FC<
...
 
         <Stack spacing="gap-y-1" align="center">
-          <Text variant="h5" align="center" selectable={false}>
-            {title}
-          </Text>
-          <Text variant="body2" customStyle="text-grey4 dark:text-grey6 my-4" selectable={false}>
-            {details}
-          </Text>
+          {typeof title === 'object' ? (
+            <>{title}</>
```
josenriagu, line 35: can we stick to having only strings as title? we can extract the error message and pass as title

libs/design-system-core/src/components/ErrorLoader/error-card.tsx
```diff
@@ -31,12 +31,20 @@ const ErrorCard: React.FC<
...
+          ) : (
+            <Text variant="h5" align="center" selectable={false}>
+              {title}
+            </Text>
+          )}
+          {typeof details === 'object' ? (
```
josenriagu, line 41: same here

libs/design-system-core/src/components/ErrorLoader/error-card.tsx
didd: This is to allow different use cases such as the one on this PR in which a title needs a line break and in such cases `ReactElement` fits.

josenriagu (approved):

PR #2430, opened 2024-10-04
chore(): show list of default extensions

didd:
## Description

The main objective of this PR is to show the list of default extensions. Other than that the following changes are introduced :
-  Move installed apps code into its own component and add a logic to check if a user is authenticated or not
- Add default route handler for profile app when the user is authenticated(created a separate ticket to handle the case when the user is not authenticated).
- Add `noWrapperCard` flag to `ErrorLoader` component to determine whether to wrap it with `Card` component or not
- Add `selectApps` selector for fetching apps by publisher DID query

## Checklist

### DIFF
--- a/extensions/apps/extensions/src/components/pages/installed-extensions/default-extensions-list.tsx
+++ b/extensions/apps/extensions/src/components/pages/installed-extensions/default-extensions-list.tsx
@@ -1,0 +1,92 @@
+import React from 'react';
+import getSDK from '@akashaorg/core-sdk';
+import AppList from '@akashaorg/design-system-components/lib/components/AppList';
+import Card from '@akashaorg/design-system-core/lib/components/Card';
+import Stack from '@akashaorg/design-system-core/lib/components/Stack';
+import Text from '@akashaorg/design-system-core/lib/components/Text';
+import Spinner from '@akashaorg/design-system-core/lib/components/Spinner';
+import Button from '@akashaorg/design-system-core/lib/components/Button';
+import ErrorLoader from '@akashaorg/design-system-core/lib/components/ErrorLoader';
+import { useRootComponentProps } from '@akashaorg/ui-awf-hooks';
+import { useGetAppsByPublisherDidQuery } from '@akashaorg/ui-awf-hooks/lib/generated';
+import { selectApps } from '@akashaorg/ui-awf-hooks/lib/selectors/get-apps-by-publisher-did-query';
+import { useTranslation } from 'react-i18next';
+import { InstalledExtension } from './use-installed-extensions';
+
+export const DefaultExtensionsList: React.FC<unknown> = () => {
+  const { t } = useTranslation('app-extensions');
+  const { getCorePlugins, worldConfig } = useRootComponentProps();
+  const sdk = getSDK();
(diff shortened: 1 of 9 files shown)


extensions/apps/extensions/src/components/pages/installed-extensions/default-extensions-list.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const sdk = getSDK();
+  const defaultApps = [worldConfig.homepageApp, ...(worldConfig.defaultApps || [])];
+
+  const { data, error, loading } = useGetAppsByPublisherDidQuery({
+    variables: {
+      id: sdk.services.gql.indexingDID,
+      filters: {
+        or: defaultApps.map(app => ({ where: { name: { equalTo: app } } })),
+      },
+      last: defaultApps.length,
```
kenshyx: you should add also sorting, otherwise the results might come in different order between calls

libs/design-system-core/src/components/ErrorLoader/index.tsx
```diff
@@ -36,7 +39,16 @@ export type ErrorLoaderProps = React.PropsWithChildren<{
  * ```
  **/
 const ErrorLoader: React.FC<ErrorLoaderProps> = ({ children, ...props }) => {
-  const { type, publicImgPath = '/images' } = props;
+  const {
+    type,
+    publicImgPath = '/images',
```
kenshyx, line 44: is this applicable to multiple components? is there a way to globally set/change this path besides specifying it as a prop?

didd: We can use the env variable for this that way we can set and change it from one place. I will create a ticket because this pattern(setting it via prop) is used in many places throughout the app and we can tackle them all at once.

kenshyx: I would say that this configuration could come from context(ex theme), having it from env is too rigid since every consumer might define a different path. Setting a default value for this is ok

extensions/apps/extensions/src/components/pages/installed-extensions/default-extensions-list.tsx
kenshyx: you could also flip the options to `first: ...` and `sorting: { createdAt: SortOrder.Desc }`, this makes the pagination easier

libs/design-system-core/src/components/ErrorLoader/index.tsx
didd: I agree. We can use theme context. I have created a ticket for this on the backlog. I will mention it in the next refinement call.

SeverS (dismissed):

extensions/apps/extensions/src/components/pages/installed-extensions/default-extensions-list.tsx
didd: I've changed last to first but the sort order we want it to be ascending so that it starts from Antenna app otherwise it will start from Vibes.

kenshyx (approved):

josenriagu (approved):

PR #2440, opened 2024-10-14
Fix: Editor word break

josenriagu:
## Description

- break word on editor rather than break all

## Checklist

### DIFF
--- a/libs/design-system-components/src/components/Editor/renderers.tsx
+++ b/libs/design-system-components/src/components/Editor/renderers.tsx
@@ -117,7 +117,10 @@ const renderElement = (
 
     default:
       return (
-        <p className={tw(`text-black dark:text-white ${textAlignStyle}`)} {...props.attributes}>
+        <p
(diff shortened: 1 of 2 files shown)


didd (dismissed):

libs/design-system-components/src/components/Editor/renderers.tsx
```diff
@@ -117,10 +117,7 @@ const renderElement = (
 
     default:
       return (
-        <p
-          className={tw(`text-black dark:text-white break-all ${textAlignStyle}`)}
-          {...props.attributes}
-        >
+        <p className={tw(`text-black dark:text-white ${textAlignStyle}`)} {...props.attributes}>
```
quininez: maybe switch to break-word instead of the default break-normal?

josenriagu: Alright, I'll attach this right away

quininez (approved):

didd (approved):

PR #2443, opened 2024-10-16
chore(): add tags validation on edit extensions form

didd:
## Description

-  add tags validation on edit extensions form

## Checklist


josenriagu (approved):

libs/design-system-components/src/components/ExtensionEditStep3Form/index.tsx
```diff
@@ -149,14 +150,16 @@ const ExtensionEditStep3Form: React.FC<ExtensionEditStep3FormProps> = props => {
...
     if (formValues.license === Licenses.OTHER) {
       formValues.license = formValues.licenseOther;
     }
+
     if (isValid) {
-      nextButton.handleClick(formValues);
+      nextButton.handleClick({
+        ...formValues,
+        contributors: addedContributors?.map(profile => profile?.did?.id),
+        keywords: [...keywords]?.filter(keyword => keyword),
```
kenshyx, line 161: is this filter to remove possible `null`, `undefined` or `false` values?

didd: Yes, thats my understanding as well.  Actually this filter was there before this PR. There is an array of available keywords the component will receive and add it to the current keywords set as the user selects one. This is one such cases where null or undefined may be part of the keywords set.

quininez (approved):

PR #2448, opened 2024-10-18
feat(notifications): add broser notification permission check

kunstefix:
## Checklist

### DIFF
--- a/libs/app-loader/src/index.ts
+++ b/libs/app-loader/src/index.ts
@@ -315,6 +315,11 @@ export default class AppLoader {
     if (!loginData?.id) {
       return;
     }
+
+    // initialise browser notifications
+    const sdk = getSDK();
+    await sdk.api.profile.initNotificationsStream();
+
     this.user = { id: loginData.id };
     this.userExtensions = await getUserInstalledExtensions();
     if (!this.userExtensions?.length) {
--- a/libs/sdk/src/profiles/index.ts
+++ b/libs/sdk/src/profiles/index.ts
@@ -157,18 +157,24 @@ class AWF_Profile {
       // account: pushAccount,
     });
 
+    const notificationsEnabled = await this.enableBrowserNotifications();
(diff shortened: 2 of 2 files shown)


libs/sdk/src/profiles/index.ts
```diff
@@ -157,18 +157,24 @@ class AWF_Profile {
       // account: pushAccount,
     });
 
+    const notificationsEnabled = await this.enableBrowserNotifications();
```
SeverS, line 160: There are cases in which 'enableBrowserNotifications' function throws an error and those cases are not handled here

kunstefix: So there are two potential points of failure:

1. Browser does not support push notifications - we need to trigger notifications through service worker (triggering them via `new Notification()` is no supported in all browsers). I will implement through existing service worker then. 

2. Client not initialised - this error is expected in case our implementation failed to create client instance. Should we use any logger here to note the bug or just let the thrown error bubble up?

kenshyx: Let's have a call tomorrow to understand better the issues you have encountered @kunstefix

libs/app-loader/src/index.ts
```diff
@@ -315,6 +315,11 @@ export default class AppLoader {
     if (!loginData?.id) {
       return;
     }
+
+    // initialise browser notifications
+    const sdk = getSDK();
+    await sdk.api.profile.initNotificationsStream();
```
kenshyx, line 321: I think this method needs to have support to specify if it should use the signer(signature request- auth push sdk) or it should just use the current eth public address(no signature - resume). Also it should treat the case where the currently signed in user does not have an eth address(ex: no supported)

kenshyx: In this specific use-case it should always try to resume and not trigger the signature request.

kunstefix, line 321: @kenshyx added initialization call here after login is made. I think thats all for the `DEV-402` now.

PR #2450, opened 2024-10-22
Fix: DS Custom Icons

josenriagu:
## Description

- replace custom app icon with hero icon
- update connect button ux
- update stories

## Checklist


libs/design-system-components/src/utils/iconByAppType.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
SeverS: iconByAppType is actually a component so no need to put it in the utils. Maybe create a generic component like ExtensionIcon and also add a story

josenriagu: Aha, yyea. Will do that now

SeverS: another question: why is this not in the ds-core?

josenriagu: Most of the components requiring it were in DS components. That's why I kept it there

SeverS: I think it will be best to move it to ds-core - adjust typings

josenriagu: alright, that works

SeverS (approved):

quininez (approved):

PR #2469, opened 2024-10-30
feature(Notifications): Add methods for handling unread notifications

kunstefix:
## Checklist

### DIFF
--- a/libs/sdk/src/common/notification/notification.ts
+++ b/libs/sdk/src/common/notification/notification.ts
@@ -177,7 +177,7 @@ class NotificationService {
       raw: true,
     };
 
-    if (!this._web3.state.address) {
+    if (!this._web3.state.address?.length) {
       return [];
     }
 
--- a/libs/sdk/src/common/notification/notification.ts
+++ b/libs/sdk/src/common/notification/notification.ts
@@ -177,6 +177,10 @@ class NotificationService {
       raw: true,
     };
 
+    if (!this._web3.state.address) {
+      return [];
+    }
+
     const localStorageKey = `${this._web3.state.address}-${this.latestSeenSidKey}`;
     const latestStoredSid = parseInt(localStorage.getItem(localStorageKey) || '0');
 
--- a/libs/sdk/src/common/notification/notification.ts
+++ b/libs/sdk/src/common/notification/notification.ts
@@ -14,6 +14,24 @@ import pino from 'pino';
 import { z } from 'zod';
 import * as notificationSchemas from './notification-schemas';
 
+interface PushOrgNotification {
+  cta: string;
+  title: string;
+  message: string;
+  icon: string;
+  url: string;
+  sid: string;
+  app: string;
+  image: string;
+  blockchain: string;
+  notification: {
+    body: string;
+    title: string;
+  };
+  secret: string;
+  isUnread: boolean;
+}
+
 @injectable()
 class NotificationService {
   private _log: pino.Logger;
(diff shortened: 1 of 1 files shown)


libs/sdk/src/common/push-protocol.ts
```diff
@@ -90,16 +109,16 @@ class PushProtocol {
...
-  async getSettingsNotification(): Promise<UserSetting[]> {
+  async getSettingsNotification(): Promise<unknown[]> {
     this.initialize();
     const subscriptions: Array<ChannelInfo> = await this._pushClient!.notification.subscriptions({
       channel: this._akashaChannelAccount,
     });
     if (!subscriptions.length) throw new Error('There is no settings for this channel');
     const channelSubscriptionInfo = subscriptions[0];
-    let userSettings: UserSetting[] = [];
+    let userSettings: unknown[] = [];
```
kunstefix: This interface didn't exist in original branch so i used unknown to temporarily fix error

libs/sdk/src/common/push-protocol.ts
```diff
@@ -9,6 +9,23 @@ import AWF_Config from './config';
...
+  url: string;
+  sid: string;
+  app: string;
+  image: string;
+  blockchain: string;
+  notification: {
+    body: string;
+    title: string;
+  };
+  secret: string;
```
kenshyx: all the props are mandatory?

kenshyx: couldn't this be used as a ReturnType from the push sdk?

libs/sdk/src/common/push-protocol.ts
```diff
@@ -19,6 +36,8 @@ class PushProtocol {
   private _pushClient?: PushAPI;
   private _notificationsStream?: PushStream;
   private _akashaChannelAccount = '0x7Ebe714066149bEef3c2f60cC89eadF50Da23C26';
+  private _unreadNotifications: PushOrgNotification[] = [];
+  public readonly lastestSeenSidKey = 'lastestSeenSidKey';
```
kenshyx: this needs to be prefixed with the current authenticated did

libs/sdk/src/common/push-protocol.ts
```diff
@@ -133,6 +152,49 @@ class PushProtocol {
...
+   */
+  async fetchUnreadNotifications() {
+    const notifications = await this.getNotifications(1, 100); // don't show more than 100 urnead notifications
+    const latestSid = localStorage.getItem(this.lastestSeenSidKey);
+    const latestSidNumber = latestSid ? parseInt(latestSid) : null;
+
+    if (latestSidNumber && !isNaN(latestSidNumber)) {
+      // Filter notifications that have a sid greater than the last seen sid
+      this._unreadNotifications = notifications.filter(
+        notification => Number(notification.sid) > latestSidNumber,
```
kenshyx: Why did you use `Number` here instead of `parseInt`?

libs/sdk/src/common/push-protocol.ts
kunstefix: The return type they have in declaration file of sdk is `any`, so I created an interface for it manually.

libs/sdk/src/common/push-protocol.ts
kunstefix: changing to parseInt.

libs/sdk/src/common/push-protocol.ts
kunstefix: added a method that prepends it

kenshyx: I think this contains also a typo, shouldn't it be `latestSeen..`?

kenshyx (approved):

PR #2474, opened 2024-11-05
extension subpages fixes

SeverS:
## Description
- fixes an error in dev subpage
- fixes virtual list on releases
- adds empty states
- adds error states

## Checklist


extensions/apps/extensions/src/components/pages/info-page/sub-pages/releases-page.tsx
```diff
@@ -56,17 +56,16 @@ export const ReleasesPage = (props: ReleasesPageProps) => {
 
   const handleLoadMoreReleases = () => {
     if (releasesReq.data?.akashaAppReleaseIndex?.edges.length < releasesCount) {
-      return;
+      releasesReq.fetchMore({
+        variables: {
+          after: releasesReq.data?.akashaAppReleaseIndex?.pageInfo.endCursor,
```
kenshyx: this will error if after is null

SeverS: hmm, but can it be null? because the if condition already guards against empty response

libs/design-system-core/src/components/ErrorLoader/index.tsx
```diff
@@ -65,6 +71,9 @@ const ErrorLoader: React.FC<ErrorLoaderProps> = ({ children, ...props }) => {
     case 'list-not-available':
       imagesrc = `${publicImgPath}/list-not-available.webp`;
       break;
+    case 'empty-list':
```
didd: ~Ideally it would make sense to have one image to represent empty list. If not since we have 'list-not-available' we can rename it to `empty-list-1` and the one you added to `empty-list-2`. That way its easier to understand we have two variations of empty list. And even rename the files names.~

We have `DefaultEmptyCard` to use for empty list.

SeverS: the idea is that list-not-available is an error (aka "Failed to fetch the list" and eventually have a retry button) whereas empty list simply means that the fetch succeded but there is nothing to show. Now that i think about it this is not actually an error so it does not belong to ErrorLoader... it's just very convenient to use it this way and not create yet a new component

didd: Yeah initially I created the `list-not-available` for errors only but for the same reasoning you described I also used it for queries which succeed but has empty data according to the design. I think we will need to revisit this, I just went through ds core and saw we have `DefaultEmptyCard` which seems to be created for empty list.

extensions/apps/extensions/src/components/pages/info-page/sub-pages/releases-page.tsx
kenshyx: I just reviewed the block here

kenshyx (dismissed):

libs/design-system-core/src/components/ErrorLoader/index.tsx
SeverS: reverted the error-loader and replaced with `DefaultEmptyCard` component

extensions/apps/extensions/src/components/pages/info-page/sub-pages/releases-page.tsx
SeverS: added guarding against possible null value of  `pageInfo.endcursor`

kenshyx (approved):

didd (approved):

PR #2476, opened 2024-11-06
Dev 412

AndiAkasha:
Parse metadata from notification & add filtering options by channel index to inbox notifications

## Checklist

### DIFF
--- a/libs/sdk/src/common/notification/notification.ts
+++ b/libs/sdk/src/common/notification/notification.ts
@@ -161,7 +161,7 @@ class NotificationService {
   )
   async getNotifications(
     page: number = 1,
-    limit: number = 100,
+    limit: number = 30,
     channelOptionIndexes: notificationSchemas.ChannelOptionIndex[] = [],
   ): Promise<notificationSchemas.PushOrgNotification[]> {
     if (!this._web3.state.address?.length) {
@@ -315,16 +315,14 @@ class NotificationService {
   }
 
   private getLatestStoredNotificationID() {
-    const localStorageKey = this.getLocalStorageKeyOfLatestSeenNotification();
-    return parseInt(localStorage.getItem(localStorageKey) || '0', 10);
+    return parseInt(localStorage.getItem(this.localStorageKeyOfLatestSeenNotification) || '0', 10);
   }
 
   private setLatestStoredNotificationID(val: string) {
-    const localStorageKey = this.getLocalStorageKeyOfLatestSeenNotification();
-    return localStorage.set(localStorageKey, val);
+    return localStorage.set(this.localStorageKeyOfLatestSeenNotification, val);
   }
 
-  private getLocalStorageKeyOfLatestSeenNotification() {
+  private get localStorageKeyOfLatestSeenNotification() {
     return `${this._web3.state.address}-${this.latestSeenNotificationIDKey}`;
   }
 
--- a/libs/sdk/src/common/notification/notification.ts
+++ b/libs/sdk/src/common/notification/notification.ts
@@ -176,13 +176,14 @@ class NotificationService {
     let notifications: (notificationSchemas.PushOrgNotification &
       notificationSchemas.AddedNotificationProps)[] = [];
 
+    const latestStoredNotificationID = this.getLatestStoredNotificationID();
     // if there are no options/apps specified.
     if (!optionsIndexes.length) {
       options.page = page;
       options.limit = limit;
       notifications = await this.notificationsClient.notification.list('INBOX', options);
       for (const notification of notifications) {
-        this.parseNotificationData(notification);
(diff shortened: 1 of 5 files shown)


libs/sdk/src/common/notification/notification.ts
```diff
@@ -338,6 +320,16 @@ class NotificationService {
     }
   }
 
+  private getLatestStoredNotificationID() {
+    const localStorageKey = `${this._web3.state.address}-${this.latestSeenNotificationIDKey}`;
```
kunstefix: We can extract this `localStorageKey` into a method so it is defined in a single place

libs/sdk/src/common/notification/notification.ts
```diff
@@ -283,7 +254,18 @@ class NotificationService {
         value: { channelIndex: metaData.channelIndex, data: metaData.data },
       });
     }
-    return notification;
+    // Mark as unread if their SID is greater than the stored SID and add properties for rendering usage
+    Object.defineProperty(notification, 'timestamp', {
+      value: new Date(notification.epoch),
+    });
+
+    const getLatestStoredNotificationID = this.getLatestStoredNotificationID();
```
kunstefix: Since `parseNotificationData` is called inside a for loop, it doesn't seem efficient to read local storage inside `getLatestStoredNotificationID` for every iteration. Better call it once and pass it as a param.

libs/sdk/src/common/notification/notification.ts
```diff
@@ -194,52 +149,84 @@ class NotificationService {
...
-      });
-
-      const isUnread = latestStoredNotificationID
-        ? notification.payload_id > latestStoredNotificationID
-        : true;
-      Object.defineProperty(notification, 'isUnread', {
-        value: isUnread,
-      });
+    let notifications: (notificationSchemas.PushOrgNotification &
+      notificationSchemas.AddedNotificationProps)[] = [];
```
kenshyx: can we make this intersection type more readable? at least to declare it as a type

libs/sdk/src/common/notification/notification.ts
```diff
@@ -194,52 +149,84 @@ class NotificationService {
...
+      return [];
+    }
+    // Fetch notifications
     const options: PushProtocol.FeedsOptions = {
       channels: [this._notificationChannelId],
       account: this._web3.state.address,
-      page,
-      limit,
+      page: 1,
+      limit: 100,
```
kenshyx: these defaults should be from the defaults set on the method params since those already have a default value
ex `page: page, limit: limit`

libs/sdk/src/common/notification/notification.ts
```diff
@@ -194,52 +149,84 @@ class NotificationService {
...
+          return (
+            parsedMetaData?.channelIndex !== undefined &&
+            optionsIndexes.includes(parsedMetaData.channelIndex)
+          );
+        });
+        notifications.push(...filteredNotifications);
+
+        // if there is no notification fetched then it means that there is no pages
+        hasMorePages = !!inboxNotifications.length;
+        options.page!++;
```
kenshyx: instead of using `!` here it is better to check before this while block if there is an `options.page` value being set

libs/sdk/src/common/notification/notification.ts
```diff
@@ -257,6 +244,68 @@ class NotificationService {
...
+
+    const isUnread = latestStoredNotificationID
+      ? notification.payload_id > latestStoredNotificationID
+      : true;
+    Object.defineProperty(notification, 'isUnread', {
+      value: isUnread,
+    });
+  }
+
+  parseMetaData(metaData: notificationSchemas.AdditionalMetadata) {
```
kenshyx: is this method required to be public or it's just internal?

libs/sdk/src/common/notification/notification.ts
```diff
@@ -257,6 +244,68 @@ class NotificationService {
     return false;
   }
 
+  parseNotificationData(
```
kenshyx: same with this one, I think it could be just and internal method

libs/typings/src/sdk/notification.ts
```diff
@@ -4,8 +4,8 @@ export enum ChannelOptionIndexes {
   VIBES = 3,
 }
 export type NotificationParsedMetaData = {
-  data: NotificationMetaTypes;
-  channelIndex: number;
+  data: NotificationMetaTypes | string;
+  channelIndex: number | undefined;
```
kenshyx: undefined means that this property is optional

kenshyx: you can use `channelIndex?: number`

kenshyx (approved):

PR #2505, opened 2024-12-03
Fix edit extension and publish release 

quininez:
## Description
* fixes for edit published extension and publish release flow
* save form state when navigating to gallery page
* handle form default image object structure and save gallery images when publishing
* prevent automatic form submit on publish release

## Checklist

### DIFF
--- a/extensions/apps/extensions/src/components/pages/extension-edit-published-page/edit-published-extension-page.tsx
+++ b/extensions/apps/extensions/src/components/pages/extension-edit-published-page/edit-published-extension-page.tsx
@@ -19,8 +19,10 @@ import { selectAppData } from '@akashaorg/ui-awf-hooks/lib/selectors/get-apps-by
 import { MAX_GALLERY_IMAGES } from '../../../constants';
 import { AtomContext, formDefaultData } from './main-page';
 import { useAtom } from 'jotai';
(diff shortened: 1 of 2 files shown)


didd (dismissed):

extensions/apps/extensions/src/components/pages/extension-edit-published-page/edit-published-extension-page.tsx
```diff
@@ -20,6 +20,16 @@ import { MAX_GALLERY_IMAGES } from '../../../constants';
 import { AtomContext, formDefaultData } from './main-page';
 import { useAtom } from 'jotai';
 
+const getImageObject = imageWithExtraProps => {
```
SeverS: can you add typings to the `imageWithExtraProps` param? because i don't understand why do you need this function

quininez: when the image data is saved from the model to the form it contains also the __typename property from graphql, this is to clean that up before performing the mutation, I can add the typings for the image object so that its clear that what is passed here should contain those

didd (approved):

SeverS (approved):

PR #2529, opened 2025-01-23
Fix: Add default false values to notifications option list

AndiAkasha:
Latest changes in Push Notification Protocol now it does not subscribe the user to the options by default and the 'getSettingsOfUser' of notificationService in sdk returns an empty array. This caused an error in the view which made impossible for the user to change the settings.

## Checklist


extensions/apps/settings/src/components/pages/notifications-preferences-option/index.tsx
```diff
@@ -50,8 +50,10 @@ const NotificationsPreferencesOption: React.FC = () => {
   useEffect(() => {
     if ((notificationsEnabled || readOnlyMode) && initialLoading) {
       sdk.services.common.notification.getSettingsOfUser().then(fetchedPreferences => {
```
kenshyx: please add error management

kenshyx (approved):

quininez (approved):

PR #2532, opened 2025-01-27
Fix editor rich text

quininez:
## Description
* when formatting text, slate doesn't update children, so force an update whenever text is formatted to work around this
* fixes an issue where after publishing the modal confirming navigation appears

## Checklist


josenriagu (approved):

libs/design-system-components/src/components/Editor/index.tsx
```diff
@@ -448,6 +448,8 @@ const EditorBox: React.FC<EditorBoxProps> = props => {
 
   const publishDisabled = publishDisabledInternal || disablePublish;
 
+  const [, forceUpdate] = React.useReducer(x => x + 1, 0);
```
SeverS, line 451: why is a forced update needed?

quininez: because when formatting text slate doesn't update the children, the force update is for the changes to propagate to the children so the content block has the latest state of the editor. And since this happens only when formatting text it is not very intensive, as it happens infrequently

SeverS (approved):

PR #2536, opened 2025-01-31
Dev 539

kunstefix:
## Description

Updated ErrorLoader component to use shadcn-based components internally. 

## Checklist


didd:
@kunstefix the discussion we had was to place the error loader inisde the ui package. Otherwise we are introducing unnecessary confusion. None of the components in the ds core will be used going forward everything that will be used is in the ui package. And when you want to use this package in a specific extension you add it to the corresponding `project.json` file like we discussed before.

didd (changes requested):
@kunstefix the discussion we had was to place the error loader inisde the ui package. Otherwise we are introducing unnecessary confusion. None of the components in the ds core will be used going forward everything that will be used is in the ui package. And when you want to use this package in a specific extension you add it to the corresponding project.json file like we discussed before.

didd:
@kunstefix I have more review items on this PR ... to unblock Vali I created another PR for the error loader component https://github.com/AKASHAorg/akasha-core/pull/2538

PR #2540, opened 2025-02-04
chore(): adjust ui package components

didd:
## Description

-  adjust ui package components

## Checklist

### DIFF
--- a/libs/ui/src/akasha-components/stepper.tsx
+++ b/libs/ui/src/akasha-components/stepper.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef } from 'react';
+import * as React from 'react';
 import { Check, Dot } from 'lucide-react';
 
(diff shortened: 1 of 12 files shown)


quininez (approved):

libs/ui/src/akasha-components/stepper.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { forwardRef } from 'react';
+import * as React from 'react';
```
kenshyx, line 1: Why do you prefer this type of import?

didd: i saw its preferred among component libraries while using Typescript to avoid **potential** import issues ... in this commit's case im keeping it consistent with how all components in shadcn are importing it

kenshyx (approved):

PR #2542, opened 2025-02-07
Fix: adjust to changes of PushProtocolApi

AndiAkasha:
The PushProtocol API does not automatically subscribe the user to all channel notifications.
When fetching the user's notification settings via `getSettingsOfUser()`, an empty array may be returned
if the user has not explicitly opted in to any notifications. In this case, we:
 1. Fetch the default channel settings using `getSettingsOfChannel()`.
 2. Mark these channel notifications as not selected (disabled) so that the user must perform an additional action to enable them.

## Checklist


extensions/apps/settings/src/components/pages/notifications-option.tsx
```diff
@@ -106,7 +109,9 @@ const NotificationsOption: React.FC = () => {
             isLoading={waitingForSignature}
             noWrapperCard={true}
             handleButtonClick={handleEnableNotifications}
-            text={t('You’ll be prompted with 1 signature')}
+            text={t(`You’ll be prompted with {{numberOfSignatures}} signature`, {
```
AndiAkasha, line 112: The current implementation of PushProtocol now requires two signatures for users that subscribe for the first time.

extensions/apps/settings/src/components/pages/notifications-option.tsx
```diff
@@ -39,7 +40,9 @@ const NotificationsOption: React.FC = () => {
   const TOAST_TEXTS = {
     success: {
       title: t('In-app notifications enabled'),
-      description: t('Notifications for all default apps are enabled. Manage them in preferences.'),
+      description: t(
+        'Notifications for all default apps are disabled. Manage them in preferences.',
```
AndiAkasha, line 45: The current implementation of 'Initialize' does not subscribe the user to all the channels by default

extensions/apps/settings/src/components/pages/notifications-preferences-option/utils.ts
```diff
@@ -1,27 +1,72 @@
...
+      index: setting.index,
+      appName: setting.description,
+      title: appInfo.title,
+      description: appInfo.description,
+      enabled: value,
+    };
+  });
+};
+
+const getAppInfoByAppName = (appName: string) => {
```
AndiAkasha: In case a new Option is added in the PushProtocol's channel this will generate a default title and description for that

extensions/apps/settings/src/components/pages/notifications-option.tsx
SeverS: you can add this type of comments to the code itself. in jsx you can use {/* Comment here */}

extensions/apps/settings/src/components/pages/notifications-option.tsx
SeverS: same as above

extensions/apps/settings/src/components/pages/notifications-preferences-option/utils.ts
SeverS: same as above

extensions/apps/settings/src/components/pages/notifications-preferences-option/utils.ts
```diff
@@ -1,27 +1,73 @@
...
-    //   appName: AppName.VIBES,
-    //   enabled: val,
-    // },
-  ];
+const appInformation = [
+  {
+    appName: AppName.ANTENNA,
+    title: 'Antenna',
+    description:
+      'Get notifications about new reflections on your beams people you follow & your interests.',
```
kunstefix: Texts should be wrapped with translation function

extensions/apps/settings/src/components/pages/notifications-preferences-option/utils.ts
```diff
@@ -1,27 +1,73 @@
...
+  });
+};
+
+const getAppInfoByAppName = (appName: string) => {
+  const appInfo = appInformation.find(appInfo => appInfo.appName === appName);
+  // In case a new Option is added in the PushProtocol's channel and we did not cover the title and the description in the appInformation array this will generate a default title and description so it be present in the list of apps
+  if (!appInfo) {
+    return {
+      title: appName,
+      description: `Get notifications from ${appName}`,
```
kunstefix: Translation function needed here as well

kunstefix (dismissed):

kunstefix (approved):

SeverS (approved):

PR #2548, opened 2025-02-11
Replace Card component

kunstefix:
## Checklist


didd (changes requested):
1. We already have the card and typography components in the ui package. We have them inside the `akasha-components`. Why create another one ?

2. The package you added to the local package.json of ui package already exists in the root package.json.

didd:
We discussed last time there was no need to spend time and energy on the vibes and legal apps. :/

didd:
@kunstefix can you convert the `px` values into `rem` ?

didd (changes requested):
1. The cookie card style is not right, please check it

2. The beam card content style has issues. The text is partially masked because of a border radius on the card wrapping the beams content. The same issue exists on the reflection cards. 
image

3. The style of the error card inside the beams needs a fix so does the install extension card. (The border radius and the gap need a fix)
image
image

4. The border radius of the beam card, the cards on reflection page, my antenna, tags feed page are too much imo. We will address this in another PR if Basma thinks what you have on your PR is fine.

5. The card style on the extension info page needs a fix so does the one on profile info page

image

image

6. The beams page on profile app needs a fix. The list container has background and the nsfw card looks weird.

image
 


7. Can you convert the px values you used on this PR into rem ?

libs/feed/src/components/beam-feed.tsx
```diff
@@ -169,7 +169,7 @@ const BeamFeed = (props: BeamFeedProps) => {
         />
       )}
       {beams && (
-        <Card ref={vListContainerRef} type="plain">
+        <Card ref={vListContainerRef} className="p-0 border-none bg-transparent">
```
didd:
```suggestion
        <div ref={vListContainerRef}>
```

libs/feed/src/components/beam-feed-by-author.tsx
```diff
@@ -164,7 +164,7 @@ const BeamFeedByAuthor = (props: BeamFeedByAuthorProps) => {
         />
       )}
       {beams.length > 0 && (
-        <Card ref={vListContainerRef} type="plain">
+        <Card ref={vListContainerRef} className="p-0 border-none bg-transparent">
```
didd:
```suggestion
        <div ref={vListContainerRef}>
```

libs/feed/src/components/beam-feed.tsx
kunstefix: Updated

libs/feed/src/components/beam-feed-by-author.tsx
kunstefix: Updated

libs/feed/src/components/beam-feed.tsx
didd: it's not updated

didd (dismissed):

josenriagu (changes requested):
Good job Klemen. Could you also check the profile header?
Screenshot 2025-02-18 at 12 54 20

didd (approved):

josenriagu (approved):

PR #2555, opened 2025-02-20
World builder

quininez:
## Description
* refactor page structure
* generate hooks for fetching a world by creator DID
* create world data flow (UI + saving/fetching data)
* create world config flow  (UI + saving/fetching data)
* dashboard stub (limited UI, fetching data)

## Checklist


libs/hooks/src/selectors/get-worlds-by-creator-did-query.ts
```diff
@@ -0,0 +1,25 @@
+import { GetWorldsByCreatorDidQuery } from '@akashaorg/typings/lib/sdk/graphql-operation-types-new';
+import { AkashaWorldEdge } from '@akashaorg/typings/lib/sdk/graphql-types-new';
+
+const isEdgeList = (
+  resp: GetWorldsByCreatorDidQuery,
+): resp is { node: { akashaWorldList: { edges: AkashaWorldEdge[] } } } => {
+  return (
+    resp?.node !== undefined &&
+    'akashaWorldList' in resp.node &&
+    resp?.node.akashaWorldList.edges !== undefined &&
```
kunstefix, line 10:
```suggestion
    !!resp?.node?.akashaWorldList?.edges &&
```

extensions/apps/world-builder/src/components/pages/world-config/world-config-form-step2.tsx
```diff
@@ -0,0 +1,358 @@
...
+        ).then(() =>
+          navigate({
+            to: '/config-success',
+            search: {
+              worldId: worldId,
+              worldName: worldData?.name,
+            },
+          }),
+        );
+      },
```
kunstefix: This would be more readable if broken down into several functions and called like this: 

```
onCompleted: async (data) => {
  const worldConfigID = data?.setAkashaWorldConfig?.document?.id;
  if (!worldConfigID) {
    showErrorNotification(t("Failed to retrieve world configuration ID."));
    return;
  }

  const extensionIDs = getUniqueExtensionIDs();
  await createExtensions(extensionIDs, worldConfigID);

  navigateToSuccessPage();
};

```

extensions/apps/world-builder/src/components/pages/world-config/world-config-form-step1.tsx
```diff
@@ -0,0 +1,179 @@
...
+                  </FormDescription>
+
+                  <Select onValueChange={field.onChange} required>
+                    <FormControl>
+                      <SelectTrigger className="w-full">
+                        <SelectValue placeholder={t('Select a layout extension')} />
+                      </SelectTrigger>
+                    </FormControl>
+                    <SelectContent>
+                      {layoutExtensionOptions?.length > 0 &&
```
kunstefix: Checking `layoutExtensionOptions?.length > 0`is redundant

kunstefix (changes requested):

libs/hooks/src/selectors/get-worlds-by-creator-did-query.ts
quininez: This is a type predicate to exclude {},  the response is something like node.akashaWorldList | {} , with that change typescript will complain 'prop x not found on {}'

kunstefix (approved):

didd (approved):

extensions/apps/world-builder/tsconfig.json
```diff
@@ -5,7 +5,11 @@
     "rootDir": "./src",
     "baseUrl": ".",
     "paths": {
-      "@/*": ["src/*"]
+      "@/*": ["src/*"],
+      "@akashaorg/core-sdk": ["../../../dist/npm/core-sdk/lib"],
```
kenshyx, line 9: these definitions are already in the base tsconfig, why did you have to declare them again?

kenshyx (changes requested):

extensions/apps/world-builder/tsconfig.json
kenshyx: also, why there is a css file(main.css) among configuration files ;/

quininez: somehow the tsconfig here was overwriting the base config, without redefining here the sdk imports were not working because of the added baseUrl

kenshyx: yes, because you have added `"@/*": ["src/*"]`

kenshyx: all the aliases should be handled in the base tsconfig

quininez: that and the baseUrl was added by Didd to allow having ui components specific to the app in the package

kenshyx: that alias should be in the base tsconfig

quininez: we tried it when we had a call about this and it didn't work, I tried it again now and it fails

didd: I see what Marius shared, we can leave the baseUrl as it is but move the alias to tsconfig base ... I have a PR here https://github.com/AKASHAorg/akasha-core/pull/2561

I added the main.css to allow adding specific themes per extension and I have moved it inside the `src` folder.

kenshyx: this will end up having duplicated configuration across all the extensions that use the ui library :/

didd: to import components from the ui library none of the this configs are needed within the world builder, they are only needed because we want to add some components directly inside extensions using the shadcn cli and enable customisation specific to the extension

in this context, the alias and baseUrl set inside the tsconfig of world-builder do two things:

1. the baseUrl tells the shadcn cli to add components inside the folder marked by baseUrl 
2. all imports inside these components use the alias, there will be no need to adjust the import path

the alias can be moved to base tsconfig but then the path for core-sdk will no longer work as baseUrl is the current directory 

i can rely on manually adding components from the registry into extensions. i can just copy paste the components, adjust the import paths and thats it. the only thing is that i wont be able to use the shadcn cli in the extensions which is fine. in this case we can remove the baseUrl, the alias and adjust the import paths, so the `components.json`, and `main.css` can be removed

didd: to import components from the ui library none of these configs are needed within the world builder, they are only needed because we want to add some components directly inside extensions using the shadcn cli and enable customisation specific to the extension

in this context, the alias and baseUrl set inside the tsconfig of world-builder do two things:

1. the baseUrl tells the shadcn cli to add components inside the folder marked by baseUrl

2. all imports inside these components use the alias, there will be no need to adjust the import paths

the alias can be moved to base tsconfig but then the path for core-sdk will no longer work as baseUrl is the current directory

i can rely on manually adding components from the registry into extensions. i can just copy paste the components, adjust the import paths and thats it. the only thing is that i wont be able to use the shadcn cli in the extensions which is fine. in this case we can remove the baseUrl, the alias and adjust the import paths, and the components.json can be removed

the main.css is used to define custom theme for the extension(i will move it inside the src folder). and a theme is added using css vars, something like this :

```
#world-builder {
   --primary: 54.1% 0.247 293;
}

#world-builder .dark {
   --primary: 69.2% 0.179 298.8;
}
```
the above theme will be specific to world builder

kenshyx (approved):

PR #2577, opened 2025-03-11
world preview

SeverS:
## Checklist


extensions/apps/world-builder/src/components/pages/main/dashboard.tsx
```diff
@@ -60,13 +60,17 @@ export const DashboardPage: React.FC = () => {
   const {
     data: worldsByCreatorDidReq,
     loading: loadingWorldsByCreatorDidQuery,
-    error: worldsByCreatorDidError,
```
quininez: the error is used now in the code, you can add it back

quininez (changes requested):

worlds/akasha.world/src/get-world-config.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const worldConfig = selectWorldConfigData(worldConfigRes);
+
+  if (!worldConfig) {
+    return null;
+  }
+
+  formattedWorldConfig.homepageApp = (
+    await getExtensionInfo(worldConfig.homepageExtension, gqlClient)
+  ).name;
```
quininez, line 70: should guard here in case fetch fails

SeverS: it will throw and caught in the caller. no point of guarding here because there is no logic, just a return statement

quininez (approved):

kunstefix (approved):

PR #2580, opened 2025-03-13
chore(): update Stepper component from akasha-ui

kunstefix:
## Checklist


libs/ui/src/akasha-components/stepper.tsx
```diff
@@ -3,21 +3,17 @@ import { Check, Dot } from 'lucide-react';
...
->(({ currentStep, numberOfSteps, className, ...props }, ref) => {
+const Stepper = ({
+  currentStep,
+  numberOfSteps,
+  className,
+  ...props
+}: React.ComponentProps<'nav'> & {
+  currentStep: number;
+  numberOfSteps: number;
+}) => {
```
didd: on akasha core we need `forwardRef` to support `ref` as we are still using react v18 ... its not needed in akasha ui because we migrated to v19 and `forwardRef` is deprecated on this version

i am curious did you remove this line after giving it thought or you just did a copy paste from akasha ui ?

kunstefix: I noticed forwardref was removed in akasha-ui compared to core. I checked component usages and noticed that ref is not assigned anywhere. I also remembered that initially we were adding forwardRef to all components and then removed it from some of them, so I assumed Stepper had it removed and and then I decided to just copy the code over.

kunstefix: added it back now

didd (approved):

quininez (approved):

PR #2581, opened 2025-03-14
chore(): various styling fixes

kunstefix:
## Checklist

### DIFF
--- a/extensions/apps/extensions/src/components/pages/my-extensions/my-extensions.tsx
+++ b/extensions/apps/extensions/src/components/pages/my-extensions/my-extensions.tsx
@@ -207,7 +207,7 @@ export const MyExtensionsPage: React.FC<unknown> = () => {
         direction="row"
         justifyContent="between"
         alignItems="center"
-        className="p-3 bg-inherit rounded-[1.25rem] bg-muted"
+        className="p-3 bg-inherit rounded-[1.25rem] bg-nested-card"
       >
         <Text variant="body1">{t('Create an extension ✨ 🚀')}</Text>
         <Button size="sm" onClick={handleNavigateToCreateApp}>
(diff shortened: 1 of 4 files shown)


extensions/widgets/layout/src/components/index.tsx
```diff
@@ -268,7 +268,7 @@ const Layout: React.FC<unknown> = () => {
                 </Stack>
               </Stack>
             </Stack>
-            <Stack className="fixed bottom-2 lg:(w-[21rem])">
+            <Stack className="fixed bottom-2 lg:(w-[21.125rem])">
```
didd, line 271: this value is weird, 21.125rem :)) ...i see what you are trying to fix but did you look for another way to fix it ?

didd, line 271: this solution looks brittle to me ... did you try another means to fix it ? you are trying to fix alignment problem between cookie widget with the rest of the widgets

kunstefix: The element is fixed and needs specified width. The previous width was too narrow that caused misalignment, so that  was the only solution I saw.

didd: ok, I will look into this when I get chance

extensions/apps/extensions/src/components/pages/my-extensions/my-extensions.tsx
```diff
@@ -207,7 +207,7 @@ export const MyExtensionsPage: React.FC<unknown> = () => {
         direction="row"
         justifyContent="between"
         alignItems="center"
-        className="p-3 bg-inherit rounded-[1.25rem]"
+        className="p-3 bg-inherit rounded-[1.25rem] bg-muted"
```
didd: isn't the color supposed to be `bg-nested-card` ?

kunstefix: bg-muted was visually closer to previous color, but since we have similar element on antenna for "Start beaming" action I changed it to bg-muted so its the same now as you suggest

didd (approved):

quininez (approved):

kenshyx (approved):

PR #2593, opened 2025-03-26
chore(): beam card conditional navigation

didd:
## Description

This PR updates the behavior of BeamCard so that navigation to the beam page is controlled by a block. For example, navigation is prevented when interacting with a poll block (to allow voting), but allowed when interacting with a slate block.

## Checklist

### DIFF
--- a/libs/hooks/src/utils/navigate.ts
+++ b/libs/hooks/src/utils/navigate.ts
@@ -1,0 +1,23 @@
+const NAVIGATE_KEY = '__navigate__';
+
+/**
+ * Sets a navigation trigger flag in localStorage.
+ * This is meant to be read once by another component,
+ * and cleared immediately after reading.
+ */
+function triggerNavigation(): void {
+  localStorage.setItem(NAVIGATE_KEY, 'true');
+}
+
+/**
+ * Checks if navigation should occur.
+ * This reads the one-time flag from localStorage and clears it after use.
+ * @returns {boolean} true if navigation was triggered; false otherwise.
+ */
+function shouldNavigate(): boolean {
+  const value = localStorage.getItem(NAVIGATE_KEY);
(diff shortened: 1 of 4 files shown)


libs/hooks/src/utils/navigate.ts
```diff
@@ -0,0 +1,23 @@
...
+  localStorage.setItem(NAVIGATE_KEY, 'true');
+}
+
+/**
+ * Checks if navigation should occur.
+ * This reads the one-time flag from localStorage and clears it after use.
+ * @returns true if navigation was triggered; false otherwise.
+ */
+function shouldNavigate(): boolean {
+  const value = localStorage.getItem(NAVIGATE_KEY);
```
kenshyx: wouldn't it be better to use sessionStorage?

libs/hooks/src/utils/navigate.ts
```diff
@@ -0,0 +1,23 @@
...
+  localStorage.removeItem(NAVIGATE_KEY);
```
kenshyx: - instead of removing the key, you can update it to "false"
- all the sessionStorage keys are removed automatically when all app the tabs are closed

libs/hooks/src/utils/navigate.ts
didd: i agree

libs/hooks/src/utils/navigate.ts
didd: i agree, will avoid the need to remove the key

libs/hooks/src/utils/navigate.ts
didd: i agree, will avoid the need to explicitly remove the key

kenshyx (approved):

josenriagu (approved):

PR #2609, opened 2025-04-25
chore(): replace checkbox component

kunstefix:
## Checklist


quininez (dismissed):

extensions/apps/extensions/src/components/pages/install-extension/install-terms-conditions.tsx
```diff
@@ -82,7 +83,7 @@ export const ExtensionInstallTerms = ({ appId }: { appId: string }) => {
     data: { authenticatedDID, isAuthenticating },
   } = useAkashaStore();
 
-  useEffect(() => {}, []);
+  useEffect(() => { }, []);
```
SeverS: why is this useEffect empty?

extensions/apps/extensions/src/components/pages/install-extension/install-terms-conditions.tsx
```diff
@@ -198,16 +198,18 @@ export const ExtensionInstallTerms = ({ appId }: { appId: string }) => {
...
-              label={fieldLabels[stateKey]}
-              handleChange={handleCheckboxChange}
-              value={'value'}
-              customStyle={'gap-x-3'}
-            />
+            <Stack direction="row" alignItems="center" spacing={2}>
+              <Checkbox
+                id={stateKey}
+                name={stateKey}
+                value={'value'}
```
josenriagu: better to assign 'stateKey' also to value so the mapped items don't share same value?

josenriagu (changes requested):

extensions/apps/settings/src/components/pages/notifications-preferences-option/app-setting.tsx
```diff
@@ -21,10 +21,20 @@ const AppSetting: React.FC<IAppSettingProps> = ({ title, description, isSelected
           id="checkbox"
           value="app-setting"
           name="app-setting"
-          isSelected={isSelected}
-          handleChange={onChange}
-          size="large"
-          customStyle="w-6 h-6"
+          checked={isSelected}
+          onCheckedChange={checked => {
+            const event = {
```
SeverS, line 26: this is too hacky. you are basically mocking an event here by providing only partial properties and passing it as it is a real event... I would adjust the onChange callback and adjust the parent

SeverS: this is too hacky. you are basically mocking an event here by providing only partial properties and passing it as it is a real event... I would adjust the onChange callback and adjust the parent

extensions/apps/settings/src/components/pages/notifications-preferences-option/enable-all-setting.tsx
```diff
@@ -22,10 +21,20 @@ const EnableAllSetting: React.FC<IEnableAllSettingProps> = ({ isSelected, onChan
           id="enable-all-notifications-checkbox"
           value="Enable all"
           name="enable-all"
-          isSelected={isSelected}
-          handleChange={onChange}
-          size="large"
-          customStyle="w-6 h-6"
+          checked={isSelected}
+          onCheckedChange={checked => {
+            const event = {
```
SeverS: same thing here...

extensions/apps/extensions/src/components/pages/install-extension/install-terms-conditions.tsx
kunstefix: no idea why that was there in the first place.. removed it now

josenriagu (dismissed):

didd (dismissed):

didd (approved):

josenriagu (approved):

PR #2617, opened 2025-05-02
Refactor: Replace hero icons with icons from lucide-react

josenriagu:
## Checklist


didd:
1. Some of the icons on the side bar aren't showing

image

2. The dimension of the icon on the Pill is not correct, along with the color

image

3. The same goes for the icon on app info page as shown below, and the style of ellipsis icon on the menu isn't right 

image

4. The text block icon on the beam editor is missing

image

5. The same text block icon is missing when you are on add blocks section

image


5. An icon is missing from the text toolbar and the color isn't correct

image

6. You can use arrow up and arrow down icons from lucide for move  up and down buttons to make the style consistent with with what we have on next. I suggest you discuss this with Basma. 

image


7. On the profile info page, the icon used for beams can be this one https://lucide.dev/icons/messages-square so that it looks consistent with whats we have on next. The same goes for the beam and reflection cards. And please take note of the dimensions of this icon on these two cards. Its smaller on next.

image

image

8. The info icon on the notification preference isn't correct 

image

If I notice more this morning I will let you know.

josenriagu:
@didd Thanks for spotting these. They are now fixed in the most recent commit.

didd (dismissed):

kunstefix (dismissed):

didd (approved):

kunstefix (approved):

PR #2623, opened 2025-05-09
chore(): replace list and menu with dropdown menu component

didd:
## Description

- replace list and menu with dropdown menu component

## Checklist

### DIFF
--- a/extensions/apps/antenna/src/components/pages/global-antenna-page/index.tsx
+++ b/extensions/apps/antenna/src/components/pages/global-antenna-page/index.tsx
@@ -60,11 +60,6 @@ const GlobalAntennaPage: React.FC<unknown> = () => {
     <Stack className="w-full">
       <Helmet helmetData={helmetData}>
         <title>{worldConfig.title}</title>
-        <style>{`
-          [data-scroll-locked][data-scroll-locked] {
-            overflow-y: scroll !important;
-          }
-        `}</style>
       </Helmet>
       <Stack spacing={4}>
         {!dismissed && (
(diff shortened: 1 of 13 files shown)


extensions/apps/antenna/src/components/pages/global-antenna-page/index.tsx
```diff
@@ -60,6 +60,11 @@ const GlobalAntennaPage: React.FC<unknown> = () => {
     <Stack className="w-full">
       <Helmet helmetData={helmetData}>
         <title>{worldConfig.title}</title>
+        <style>{`
+          [data-scroll-locked][data-scroll-locked] {
```
SeverS: i'm not happy with this approach:
1. it overflows the context of an app - other apps will get this style without needing it
2. it is hard to track which app is responsible of adding this style

are you using a library that doesn't have an option to style this particular state?

didd: the radix ui dropdown menu implementation removes the window scroll bar when the drop down is active which from the doc I didn't find a way to solve it that I went with introducing that style to global antennna

but now i went through their github and  figured i can turn off the modal prop of the drop down menu which helps to prevent the logic that removes the window scroll bar but on the other side now when the modal is active one can interact with other things beside the modal which assessed is fine

didd: the radix ui dropdown menu implementation removes the window scroll bar when the drop down is active which looking at their doc I didn't find a way to solve it that I went with introducing that style only to global antennna

but now i went through their github and figured i can set the modal prop of drop down menu to false which helps to prevent the logic that removes the window scroll bar but on the other side now when the modal is active you can interact with the surrounding components which is fine ... modal prop is set to true by default which prevents interaction with outside element and also removes the window scrollbar

SeverS (approved):

quininez (approved):

